Issue #12, opened 2019-11-12
`Makefile` issues


BatmanAoD:
I found two issues with the `Makefile`:

* It fails if the `docs` branch has not previously been checked out locally. This can be easily fixed by using `origin/docs`, but I can imagine that having some surprising behavior for the maintainer when actually trying to modify the docs. `git branch -u origin/docs docs` [_should_](https://stackoverflow.com/a/51157918/1858225) ensure the branch exists locally without actually changing the working directory, but for some reason that isn't working on my machine so I'm hesitant to open a PR with that change.
* The `mkdir -p` commands for the documentation directory don't actually create the leaf dirs, `man` and `html`. This is a pretty simple fix.

If you're okay with using `git show origin/docs...`, or if I figure out why `git branch -u` isn't working on my machine, I can submit a simple PR fixing these issues.

BatmanAoD:
I got `git branch` to do the right thing without the `-u` flag. PR opened.

BatmanAoD:
These issues will be addressed by #10.

PR #10, opened 2019-11-08
Installation fixes (some may be Windows-specific, some not)

benblo:
- documentation fixes
- fix for paths with spaces
- python shebang fix, which you may not want, and might just be an issue with my own installation (I can only use "#!/usr/bin/env python", not "python3")

Side-note:
here's what ended up working for me to install on Windows:
  git clone -c core.symlinks=true https://github.com/newren/git-filter-repo.git
  make prefix="C:\Program Files\Git\mingw64" pythondir="C:\Python38\Lib\site-packages" mandir="C:\gnuwin32\share\man" install

- The git path I believe is standard for a Git for Windows install
- Python path is from a fresh chocolatey install
- gnuwin32: not sure if it was installed by Git for Windows, or by me from earlier...

PR #13, opened 2019-11-12
Fixes for #12

BatmanAoD:
Create 'docs' branch if necessary; create Documentation subdirs

Addresses concerns in #12

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,12 +26,17 @@ fixup_locale:
 snag_docs: Documentation/man1/git-filter-repo.1 Documentation/html/git-filter-repo.html
 
 Documentation/man1/git-filter-repo.1:
-	mkdir -p man1
-	git show docs:man1/git-filter-repo.1 >Documentation/man1/git-filter-repo.1
+	mkdir -p Documentation/man1
+	# Create the local `docs` tracking branch if it doesn't already exist.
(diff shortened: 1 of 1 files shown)


newren:
Hi, thanks for the fixes.  I really need to get some contribution guidelines up.

This turns out to be a duplicate of #10; most of my comments there also apply here.  Since @benblo filed first, I'll merge his changes if he updates them in the next few days; if not, and you update yours, then I'll go ahead and merge yours.

newren:
Oh, one more thing.  I'd prefer to avoid things like "Resolves #12", as it makes the history depend on an external service that isn't obvious.  Commit messages should explain the problem in sufficient detail to understand; if that isn't possible for some reason or you want to be able to point users at additional discussion, please use a full URL.

BatmanAoD:
Oops, I did briefly search open issues but didn't see that one. I'll close this. 

"Resolves" only modifies the issue thread, not the commit history, correct?

newren:
Sorry about "Resolves"; I for some reason thought it was included in the commit message rather than just the PR summary.  Having it in the PR summary would be fine since that'd be GitHub specific anyway.

BatmanAoD:
Fair enough; that feature does feel pretty "magic" so I wanted to be sure I understood its effect as well.

PR #15, opened 2019-11-13
Mention INSTALL in README

busunkim96:


newren:
Thanks for the contribution.

I don't like the idea of adding a reference in the "how do I use it?" section, though.  Perhaps we could make an "Installation" section, and move part of the opening paragraph into that.

Also:
  * this patch records an invalid email address for you; that should be corrected.
  * missing Signed-off-by line as per https://github.com/git/git/blob/master/Documentation/SubmittingPatches#L293
  * for consistency, the commit summary should be "filter-repo (README): mention INSTALL"

Do you want to try your hand at fixing these up, or would you like me to go ahead and restructure the README a bit to make a separate installation section?

newren:
So, after reviewing the other PRs, I noticed I was repeating myself a lot, so I wrote up some contribution guidelines.  One of the other PRs pointed out I wasn't clear about which version(s) of python were required.  And others pointed out that the installation instructions on Windows weren't detailed enough (python3 is often named "python" there instead of "python3" for some reason).  And while at it I made a few other changes, including a separate installation section in the README as a side effect.

Since that bunch of changes makes this PR obsolete, I'll go ahead and close it out, but I want to thank you for taking the time to use filter-repo, notice an issue, and flag it to my attention.  You've helped make filter-repo better; much appreciated!

Issue #11, opened 2019-11-11
Upload to PyPI


Julian:
Hi. Would you consider uploading this to PyPI?

Thanks.

newren:
I don't know what this entails.  I started poking around and found https://packaging.python.org/tutorials/packaging-projects/, which didn't show how to work with just a single file.  https://stackoverflow.com/questions/12461603/setting-up-setup-py-for-packaging-of-a-single-py-file-and-a-single-data-file-wi suggests you *must* restructure the project to make a subdirectory, which is ridiculous and made me almost quit and just say no way.  http://www.agapow.net/programming/python/setuptools-and-the-single-file/ suggests there is a way around it and provides some pointers, but also notes that "The non-functional rst2beamer sat on PyPi for months before someone informed me it wasn't working", which is rather worrisome.  Do I want to go through the pain of figuring out PyPI uploading just to have a similar fate?

Also, how does PyPI handle different installations having the python3 executable named "python3" in some places and "python" in others?  That seems to be causing problems, and I'm worried uploading to PyPI is going to reinforce or extend them.

I'm not in principle against uploading to PyPI, but someone would need to do a little legwork.  If it's me, it might be a while.

Julian:
Nah you don't need to restructure the project. I'm happy to send a PR adding the `setup.py/cfg`.

I won't throw any shade on anyone who didn't find it given Python's history of subpar packaging documentation (which overall is getting better), but the solution for single files is the `pymodules` argument, which yeah like I said happy to send in a PR.

> Also, how does PyPI handle different installations having the python3 executable named "python3" in some places and "python" in others? That seems to be causing problems, and I'm worried uploading to PyPI is going to reinforce or extend them.

In general you should assume `python` always means Python 2, and `python3` always means Python 3 (which is the python-dev recommendation. Arch is weird essentially, and one of only 2 places really that the above doesn't hold true, but that also means Arch is familiar with how to unbreak itself for everyone following the upstream recommendation). If the script's written for Py3, it should have `/usr/bin/env python3` as its shebang, but it doesn't matter much, since `pip`, which is how you install things from `PyPI`, will transparently rewrite that shebang if needed when it installs the package (and will point it at the right Python).

PR incoming whenever I get some spare moments...

newren:
[...]

PR #16, opened 2019-11-16
Add packaging scripts for uploading to PyPI.

Julian:
Closes: #11.

After including these, you'd upload to PyPI by running e.g. `python3 setup.py sdist bdist_wheel && twine upload dist/*`.

(You'd need [twine](https://pypi.org/project/twine/) installed, so if you don't already have it, first e.g. `python3 -m venv venv && venv/bin/python -m pip install twine`)

### DIFF
--- a/release/setup.cfg
+++ b/release/setup.cfg
@@ -1,0 +1,29 @@
+[metadata]
+name = git-filter-repo
+url = https://github.com/newren/git-filter-repo
+project_urls =
+    Source = https://github.com/newren/git-filter-repo
+    Issues = https://github.com/newren/git-filter-repo/issues/
+description = Quickly rewrite git repository history
+long_description = file: README.md
+classifiers =
+    Development Status :: 4 - Beta
+    Operating System :: OS Independent
+    Programming Language :: Python
+    License :: OSI Approved :: MIT License
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: Implementation :: CPython
+    Programming Language :: Python :: Implementation :: PyPy
+
+[options]
+scripts = git-filter-repo
+py_modules = git_filter_repo
+python_requires = >= 3.5
+setup_requires = setuptools_scm
+
+[bdist_wheel]
+universal = 1
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,9 @@ release: update_docs
 	cat asset_id | xargs -I ASSET_ID curl -s -H "Authorization: token $(GITHUB_COM_TOKEN)" -H "Content-Type: application/octet-stream" --data-binary @$(FILEBASE).tar.xz https://uploads.github.com/repos/newren/git-filter-repo/releases/ASSET_ID/assets?name=$(FILEBASE).tar.xz
 	# Remove temporary file(s)
 	rm asset_id
+	# Upload to PyPI, automatically picking up the new tag
+	cd release && python3 setup.py sdist bdist_wheel
+	twine upload release/dist/*
 	# Notify of completion
 	@echo
 	@echo === filter-repo $(TAGNAME) created and uploaded to GitHub ===
--- a/release/git-filter-repo
+++ b/release/git-filter-repo
@@ -1,0 +1,1 @@
+../git-filter-repo
--- a/release/git_filter_repo.py
+++ b/release/git_filter_repo.py
@@ -1,0 +1,1 @@
+../git_filter_repo.py
--- a/release/pyproject.toml
+++ b/release/pyproject.toml
@@ -1,0 +1,9 @@
+[build-system]
+requires = [
+    # The minimum setuptools version is specific to the PEP 517 backend,
+    # and may be stricter than the version required in `setup.py`
+    "setuptools>=40.6.0",
+    "setuptools_scm",
+    "wheel",
+]
+build-backend = "setuptools.build_meta"
--- a/release/setup.py
+++ b/release/setup.py
@@ -1,0 +1,2 @@
+from setuptools import setup
+setup(use_scm_version=dict(root="..", relative_to=__file__))


setup.cfg
```diff
@@ -0,0 +1,27 @@
+[metadata]
+name = git-filter-repo
+url = https://github.com/newren/git-filter-repo
+project_urls = 
+    Source = https://github.com/newren/git-filter-repo
+    Issues = https://github.com/newren/git-filter-repo/issues/
+description = git filter-branch replacement
+version = 2.25.0
```
Julian: I'm assuming this is the next version you'll upload.

newren: I don't like hardcoding the version in one (or more) of the files in the repo; can we auto-generate this instead based on the current tag?

setup.cfg
```diff
@@ -0,0 +1,27 @@
...
+long_description = file: README.md
+author = Elijah Newren
+classifiers =
+    Development Status :: 3 - Alpha
```
newren: In my experience, alpha/beta/production are developer-declared fuzzy this-is-what-i-feel-it-is declarations; does PyPI have more rigid requirements on these terms?  If it doesn't, I disagree with Alpha (at least on Linux/MacOSX; on windows Alpha may be fair.)

setup.cfg
```diff
@@ -0,0 +1,27 @@
...
+    Operating System :: OS Independent
+    Programming Language :: Python
+    License :: OSI Approved :: MIT License
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
```
newren: What's the typical use and meaning here?  Do folks list the oldest known working version of python and all newer versions?  Does each version listed mean it has been tested with that version?

Most of my testing has been on 3.5.  I used 3.4 for a little while on one box, and have 3.6 on some boxes.  I haven't actually tested newer versions, though I don't have any reason to believe they would have any issues.

setup.cfg
```diff
@@ -0,0 +1,27 @@
...
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: Implementation :: CPython
+    Programming Language :: Python :: Implementation :: PyPy
```
newren: I haven't tested with PyPy yet, so I'm hesitant to claim it works with it.  Or is this not a claim that it's tested, just that it should theoretically work with it?

setup.cfg
Julian: Yep, can do that.

setup.cfg
Julian: It's just fuzzy what I feel yeah :)

Alpha is just what my packaging script emits by default, but you can choose whatever you "what-I-feel" -- what would you like? Mature?

setup.cfg
Julian: "Typically" is a tricky question :) -- small numbers of these change some things on PyPI.org -- but, roughly, what they usually mean is "this is a supported version".

Often "supported" corresponds with "tested", but not always.

Authors have different takes on what to put there -- I can tell you me as a maintainer, I generally run the test suite under all the versions I put here, but that that's over abundance of precaution -- in general, if you support some version, I think it's still safe to put all later versions on here.

(It's a quite minor concern though, these classifiers are purely informational, and you can imagine users therefore don't put a lot of implicit trust in them)

But based on what you said about your testing I'd probably personally recommend putting 3.5 and all later versions on.

setup.cfg
Julian: Uh, sorry, yeah, that's just another thing my package script autogenerates -- I think the same applies as the above, which is "if you don't use any CPython C extensions in obscene ways, it probably works on PyPy and is safe to put this here", so it's the same "supported" vs "tested" question --

I.e., if someone tried to run this tomorrow on PyPy and found some bug, would you want to investigate it or say "nah that's PyPy and not supported even to investigate" -- if the former, would probably keep this, if the latter, would axe it.

Julian:
> I would expect 'make release' would run the necessary code rather than needing to run some separate step. That could probably be added separate to this PR, but thought I'd mention it for context in case it affects how you respond to any of my other comments.

I didn't notice your Makefile, but yeah sure, if that's how you cut releases today I can put the requisite commands in there. Sounds good.

> I'm kind of sad to see three new files added to the toplevel directory. I hesitated adding INSTALL and Makefile because I didn't want another file there. I even hesitated on README.md. And I was annoyed at having three COPYING* files, though I wanted it to be clear that which text was verbatim copied from elsewhere. Is there a reasonable way to avoid this, or should I just give up already on my simple "You only need one file to make use of this new tool" pitch?

Technically speaking... not all 3 are "required"... a minimum of 2 are required for "correctness", though if you leave one out (pyproject.toml) stuff will still work today (but maybe not in X years from now), and if you leave another out (setup.cfg) you sometimes get tempted into antipatterns (setup.py is just normal Python code, and so you can do arbitrarily terrible programming things in it instead of just specifying your build process as plain static data, which is why setup.cfg exists -- but if 2 is really better than 3, in theory I can move all the stuff from setup.cfg into setup.py again)

It doesn't need to live top-level though too -- if it seems clearer to you I could move them into a `python` subdirectory or something?

Julian:
Uh on the makefile though, are you OK with me assuming that you install `twine` once globally and not install it as part of every `make release`?

setup.cfg
newren: Yeah, that's a tougher call.  If you look at it from the known issues angle (Alpha software has lots of known issues, Beta has some, for Mature it'd be rare), then it'd be on the Mature side as I've squashed the known issues.  If you look at it from the well-tested angle, then although I've done lots of testing and have line coverage and know many people who have used it including some in production and including it in a library and passing along to their users, I still know there might be a tidal wave of in-the-wild testing to come since git-2.24.0 was recently released with pointers to filter-repo.  Who knows what all that real-life testing will turn up.

So, I dunno...I guess Beta for now?

setup.cfg
newren: Ah, that makes sense -- intent to support.  So, yeah, let's keep PyPy and python >= 3.5.

newren:
> Technically speaking... not all 3 are "required"... a minimum of 2 are required for "correctness", though if you leave one out (pyproject.toml) stuff will still work today (but maybe not in X years from now), and if you leave another out (setup.cfg) you sometimes get tempted into antipatterns (setup.py is just normal Python code, and so you can do arbitrarily terrible programming things in it instead of just specifying your build process as plain static data, which is why setup.cfg exists -- but if 2 is really better than 3, in theory I can move all the stuff from setup.cfg into setup.py again)
> 
> It doesn't need to live top-level though too -- if it seems clearer to you I could move them into a `python` subdirectory or something?

Maybe a directory called release/ ?  (And perhaps I can split some code out of the Makefile and stick it into that directory as well.)

Also, in response to your other question, yeah go ahead and assume that I install twine once globally; no need to do that as part of every `make release`.

Julian:
OK, think this is ready for another review.

release/setup.cfg
```diff
@@ -0,0 +1,29 @@
+[metadata]
+name = git-filter-repo
+url = https://github.com/newren/git-filter-repo
+project_urls =
+    Source = https://github.com/newren/git-filter-repo
+    Issues = https://github.com/newren/git-filter-repo/issues/
+description = git filter-branch replacement
+long_description = file: README.md
+author = Elijah Newren
```
newren: While I've written virtually everything right now, part of the point of being open source is to allow others to contribute so I'm worried this will go stale.  Is an author field necessary?  What do larger multi-author projects do?

release/setup.cfg
```diff
@@ -0,0 +1,29 @@
+[metadata]
+name = git-filter-repo
+url = https://github.com/newren/git-filter-repo
+project_urls =
+    Source = https://github.com/newren/git-filter-repo
+    Issues = https://github.com/newren/git-filter-repo/issues/
+description = git filter-branch replacement
```
newren: Hmm...I listed that up-front on the GitHub page because it very quickly explains to people what it does.  But it's a description with an expiration date; I expect people will eventually not be familiar with filter-branch anymore.  And I'm not sure if it makes as much sense for the PyPI audience.  Maybe we should use something more generic such as "Quickly rewrite git repository history"?

release/setup.cfg
```diff
@@ -0,0 +1,29 @@
...
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: Implementation :: CPython
+    Programming Language :: Python :: Implementation :: PyPy
+
+[options]
+scripts = ../git-filter-repo
```
newren: So, I presume people using PyPI would probably be more likely than average to write a python filtering program that imported filter-repo as a module.  As far as I can tell, you are only distributing the git-filter-repo script here, but importing a python module in a file named with hyphens and without the ".py" extension is quite painful.  What about the git_filter_repo.py symlink?

glensc:
@newren:

Don't link to branches!
- see https://news.ycombinator.com/item?id=8046710
  and https://help.github.com/articles/getting-permanent-links-to-files/
- hit `y`
- and perhaps https://github.com/josephfrazier/octopermalinker

newren:
> @newren:
> 
> Don't link to branches!

@glensc: I'm confused; where in this pull request was that done?

newren:
Due to #20, I updated the [installation instructions](https://github.com/newren/git-filter-repo/blob/3e649165b5bcb9ce85d5e236e95824de624f3736/INSTALL.md) to mention package managers.  I also included a link for pip, which currently says it is coming soon and references this PR.

Julian:
> Due to #20, I updated the installation instructions to mention package managers. I also included a link for pip, which currently says it is coming soon and references this PR.

Cool!

release/setup.cfg
Julian: Yes! Good catch -- I didn't notice that `git_filter_repo.py` was a module and not just a script, but looks like you've got a `if __name__` block in it so yeah will add that to what's installed.

release/setup.cfg
Julian: You get a warning if you specify nothing -- I *think* the real PyPI may reject the package, but not sure -- the test PyPI doesn't.

In theory there's also `maintainer` instead of `author`.

In practice there are varying approaches, but the main common one is to list something to the effect of "main author and others" and then if need be link to a contributor's file.

E.g. here's what [`coverage.py` does](https://pypi.org/project/coverage/) (in the meta link in the left sidebar).

Julian:
Also FWIW as of.. [yesterday, in the next release of `setuptools-scm`](https://github.com/pypa/setuptools_scm/pull/364) you will only need 2 and not 3 files now :)

glensc:
@newren here:
- https://github.com/newren/git-filter-repo/pull/16#pullrequestreview-320242716

> As per git.git guidelines, remove the full stop at the end of the commit summary (git/git:Documentation/SubmittingPatches@master#L106-L108)
> As per git.git guidelines, please provide a Signed-off-by on your commit message (git/git:Documentation/SubmittingPatches@master#L293)

Julian:
Not sure how useful that is to mention :) -- it's for directly showing someone (me) a particular line at a particular point in time, not something that's permanently going somewhere and can drift out of sync.

newren:
> Also FWIW as of.. [yesterday, in the next release of `setuptools-scm`](https://github.com/pypa/setuptools_scm/pull/364) you will only need 2 and not 3 files now :)

Cool!

Julian:
> Before I merge, a few questions on usage (sorry for all the newbie questions; I have only used pip a few times and usually get python packages from yum/dnf install):

No worries!

> If I checkout v2.23.0, and add in these files and run cd release && python3 setup.py sdist bdist_wheel && twine upload dist/*, will it notice that I have version v2.23.0 checked out and upload a 2.23.0 version, or will it notice that the most recent tag is v2.24.0 and try to mark it with that version?

The files didn't exist back then, but assuming you copy them in, yeah this will work fine IIRC. If you do the two steps separately you'll see what version gets picked up -- and even more specifically, you can run `python setup.py --version` and it will tell you what version it thinks this is without building a distribution.

You also may want to test this out first on the [test PyPI](https://packaging.python.org/guides/using-testpypi/) so you get a feel for things before sending it out, because...

> If my upload is bad, is the only remedy to upload a newer version? Or can the existing version be re-uploaded if we find there was a problem?

Correct -- you can never mutate an existing release on PyPI -- you can only push a new one after bumping the version.

> Should I upload both 2.23.0 and 2.24.0 to PyPI, or does PyPI only make the latest version available?

It exposes all uploaded versions -- as for whether you should or not -- my recommendation as a maintainer would be "yes if you know those two versions are widely used [e.g. because debian packages your 2.23.0], but no need otherwise".

Happy to answer any other questions too.

newren:
Sorry for the delay; your update came in on US Thanksgiving.  I was planning to get this all out yesterday, but between being sick and running into a few issues, I'm still not there.

Here's what I ran:
```cd floss/git-filter-repo/release/
python3 -m venv venv
venv/bin/pip3 install --upgrade setuptools pip
venv/bin/pip3 install twine wheel
venv/bin/python3 setup.py sdist bdist_wheel
```

and then I took a look at the stuff under dist/.  In particular, `unzip -c dist/git_filter_repo-2.23.0-py2.py3-none-any.whl  | less` looked okay other than a couple fields in the METADATA file.  When I untarred dist/git-filter-repo-2.23.0.tar.gz, the files I saw looked okay (to my untrained eye) other than PKG-INFO.  That file, similar to the METADATA file in the other archive had multiple fields with a value of "UNKNOWN" -- License, Platforms, and Description.  I fixed the first two by adding the lines
```
platforms = any
license = MIT
```
to the setup.cfg file, but the Description threw me for a loop.  After digging for a while, I changed the long_description line into two lines reading
```
long-description = file: ../README.md
long_description_content_type = 'text/markdown'
```

but that failed with a Traceback, reporting
```
distutils.errors.DistutilsOptionError: `file:` directive can not access /home/newren/floss/git-filter-repo/README.md
```
That seems odd, since that is the correct full path to the filename.  A google search on the error showed that the code checked if the filename was within os.getcwd(), which I guess is where the failure comes from since cwd is the release subdirectory.  I haven't figured out how to get past that point.  Any ideas?

newren:
Okay, so adding:
```long_description_content_type = 'text/markdown'
platforms = any
license = MIT
```

to setup.cfg and ```import shutil
shutil.copy("../README.md", ".")
```

to setup.py (before the call to setuptools.setup()) while leaving the long_description field as you had it seems to fix things for me.  Do these look good to you?  Any idea why it needs separate platforms and license toplevel fields when there are already Operating System and License fields within classifiers?

Julian:
Sounds suspicious (like that I broke something while making a modification
kind of suspicious) but will take me another day or two to have a chance to
investigate.

On the classifier question -- basically classifiers do nothing, they're
opaque to pip (really setuptools), so pip still needs parameters like
license specified separately.

Yup it sucks.

Julian:
OK I didn't have a full moment to confirm 100% (will see if I can but have just started vacation), but I think actually what was happening was setuptools being unhappy with me specifying the source files relatively to the setup.cfg (having a subdirectory with the setup.py is uncommon, I can't really remember doing this before, but I was suspicious when it seemed to work :) -- the easiest way around that was to just symlink the files and not have them be relative (have a look at the new commit)

There still may be another way to do this, but yeah I think at least now the built distributions are correct.

newren:
I tweaked the commit message to add 'release:' as the area (and added my Signed-off-by since I edited it).  I also added another commit on top to fix a few issues (make sure the README.md actually gets included so that Description isn't "UNKNOWN", set the long_description_content_type so that it is parsed corrected, set the platform and LICENSE fields so they won't be "UNKNOWN", and a few others).  I then merged this in an pushed it up, and also uploaded git-filter-repo-2.23.0 and git-filter-repo-2.24.0 to PyPI.  Since git.git has done a 2.24.1 release, we can do a git-filter-repo-2.24.1 release and upload it to PyPI to correct any problems, but I think those versions should be working.

newren:
Thanks for patch and all the explanations, @Julian!

Julian:
Wow apologies, I apparently resolved this off my to-do list so completely
forgot about that last issue, but glad you got it figured out and yeah
happy to help, thanks for building this!

PR #29, opened 2019-12-11
Fix translation of mailmap entries with no email address

xorangekiller:
This PR adds a `.gitignore` file and fixes two bugs.

1. The mailmap format parsed by the `git shortlog` command allows for matching mailmap entries with no email address. This is admittedly an edge case, because most Git commits will have an email address associated with them as well as a name, but technically the address isn't required, and `git shortlog` accommodates that in its mailmap format. This PR teaches `git-filter-repo` to do the same thing, fixing what was arguably a bug (or at least a missing feature) in `git-filter-repo`.

2. I ran into this bug by accident while I was investigating the previous one. Prior to the changes made in this PR, `git-filter-repo` could only be used with either the `--dry-run` flag or the `--debug` flag, not both. When run in debug mode, `git-filter-repo` expected to be able to read from the output stream, which obviously isn't created when doing a dry run, so it stack traced when it tried to use the non-existent output stream. This PR fixes that bug with an equally simple sanity check for the existence of the output stream when run in debug mode.

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -267,7 +267,7 @@ class MailmapInfo(object):
     self._parse_file(filename)
 
   def _parse_file(self, filename):
-    name_and_email_re = re.compile(br'(.*?)\s*<([^>]+)>\s*')
+    name_and_email_re = re.compile(br'(.*?)\s*<([^>]*)>\s*')
     comment_re = re.compile(br'\s*#.*')
     if not os.access(filename, os.R_OK):
       raise SystemExit(_("Cannot read %s") % decode(filename))
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -3562,7 +3562,7 @@ class RepoFilter(object):
       self._output = open(self._fe_filt, 'bw')
     else:
       self._output = self._fip.stdin
-    if self._args.debug:
+    if self._args.debug and not self._args.dry_run:
       self._output = DualFileWriter(self._fip.stdin, self._output)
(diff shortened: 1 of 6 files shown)


git-filter-repo
```diff
@@ -3542,7 +3542,7 @@ class RepoFilter(object):
       self._output = open(self._fe_filt, 'bw')
     else:
       self._output = self._fip.stdin
-    if self._args.debug:
+    if self._args.debug and self._fip:
```
newren: I would rather this check 'not self._args.dry_run' rather than 'self._fip'; while both may be valid, I think the former documents the intent a bit better to future readers.

newren:
Oh, and sorry for taking so long to review your changes; things were busy...

git-filter-repo
xorangekiller: No problem. I see how that makes the intent a little easier to follow. I rebased my branch on the latest `master`, made that one small change, and verified that all of the tests still pass.

xorangekiller:
> Oh, and sorry for taking so long to review your changes; things were busy...

You didn't need to apologize for taking too long, but thank you. Things have been a bit crazy for me too recently with Thanksgiving (in the US), Christmas, and everything that I needed to get done around them. Thank you for taking the time to review my changes, and for writing such an awesome tool as `git-filter-repo` in the first place! It helped me recently to fix a few things in history after my conversion of a very old and very large CVS repo to Git with `cvs2git`.

newren:
Thanks for your contributions!

PR #33, opened 2019-12-22
Add callback data to be passed down to the callbacks

### LINKED ISSUE #31: Please allow to pass user_data to the callbacks
Hi,

I'm trying to write a tool to add "Reviewed-By" to commits. It would be great to have user_data which I could pass down to the commit_callback.

Thanks!

cryptomilk:
Fixes #31

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -884,7 +884,7 @@ class FastExportParser(object):
                tag_callback = None,   commit_callback = None,
                blob_callback = None,  progress_callback = None,
                reset_callback = None, checkpoint_callback = None,
-               done_callback = None):
+               done_callback = None,  callback_data = None):
     # Members below simply store callback functions for the various git
     # elements
     self._tag_callback        = tag_callback
@@ -894,6 +894,8 @@ class FastExportParser(object):
     self._progress_callback   = progress_callback
     self._checkpoint_callback = checkpoint_callback
     self._done_callback       = done_callback
+
+    self._callback_data       = callback_data
 
     # Keep track of which refs appear from the export, and which make it to
     # the import (pruning of empty commits, renaming of refs, and creating
@@ -1114,7 +1116,7 @@ class FastExportParser(object):
 
     # Call any user callback to allow them to use/modify the blob
     if self._blob_callback:
-      self._blob_callback(blob)
+      self._blob_callback(blob, self._callback_data)
 
     # Now print the resulting blob
     if not blob.dumped:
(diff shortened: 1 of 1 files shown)


git-filter-repo
```diff
@@ -1246,7 +1248,7 @@ class FastExportParser(object):
     aux_info = {'orig_parents': orig_parents,
                 'had_file_changes': had_file_changes}
     if self._commit_callback:
-      self._commit_callback(commit, aux_info)
+      self._commit_callback(commit, aux_info, self._callback_data)
```
newren: If we were to do this, I'd prefer that self._callback_data consistently be the second argument, and aux_info can be last.

git-filter-repo
```diff
@@ -1406,7 +1408,7 @@ class FastExportParser(object):
         self._parse_literal_command()
       elif self._currentline.startswith(b'done'):
         if self._done_callback:
-          self._done_callback()
+          self._done_callback(self._callback_data)
```
newren: Oh, ick, this can't be the second argument because there isn't a first argument because there's nothing to operate on.  Hmm...also, does it even make sense for the user data in the done callback?  I'm not sure, because it's just such a rare callback to want or need.

git-filter-repo
```diff
@@ -3159,9 +3166,9 @@ class RepoFilter(object):
         blob.data = regex.sub(replacement, blob.data)
 
     if self._blob_callback:
-      self._blob_callback(blob, self.callback_metadata())
+      self._blob_callback(blob, self.callback_metadata(), self._callback_data)
 
-  def _tweak_commit(self, commit, aux_info):
+  def _tweak_commit(self, commit, aux_info, callback_data):
```
newren: Again, second argument; keep aux_info last.  But more importantly, why are you passing callback_data to this function, then ignoring it and using self._callback_data anyway?  What's the point in taking it as an extra parameter?

git-filter-repo
```diff
@@ -3580,9 +3587,9 @@ class RepoFilter(object):
       print("        longer be related; consider re-pushing it elsewhere.")
     subproc.call('git remote rm origin'.split(), cwd=target_working_dir)
 
-  def _final_commands(self):
+  def _final_commands(self, callback_data):
```
newren: Same issue as above; you are passing an extra callback_data parameter, but then ignoring it and using self._callback_data.  What's the reason for passing it?  (Was this leftover from a previous attempting at passing things around and just didn't get cleaned up, by chance?)

cryptomilk:
> People can create a class and a function within a class, and then use everything in self for extra callback data (see e.g. contrib/filter-repo-demos/filter-ignore, which passes commit_callback=checker.skip_ignores)

I'm a C developer and a total python newbie. I didn't know that this works. I guess then we can close this request and I wlll rewrite my script.

newren:
Okay, I'll go ahead and close.  Thanks for your interest and submissions!

PR #34, opened 2019-12-22
Fix GPL license

cryptomilk:
The license file is outdated pointing to an incorrect FSF address.


newren:
Indeed, I had copied it from git.git into the predecessor of this project back in 2009, and git.git updated their copy in 2010 but I hadn't noticed.  Thanks for spotting and updating!

PR #55, opened 2020-02-11
setup.py: test for FileExistsError on symlink

siriobalmelli:
Multiple runs of setuptools encounter a FileExistsError exception
trying to re-symlink the same files.

This exception is safe to ignore: the files were already symlinked
so the call can be considered successful.

Signed-off-by: Sirio Balmelli <sirio@b-ad.ch>


newren:
Thanks!

Issue #98, opened 2020-05-31
Replacement for BFG's object-id-map.old-new.txt


TobiX:
Hi,

first, thanks for the awesome tool!

A question: Would it be possible to provide a replacement for BFG's `object-id-map.old-new.txt`, which just seems to be a complete map from all old blob IDs to all new blob IDs?

This is useful, for example, to fixup internal structures in GitLab after a repository rewrite: https://gitlab.com/help/user/project/repository/reducing_the_repo_size_using_git.md#using-the-bfg-repo-cleaner

jamesramsay:
I started looking at updating this documentation today and realized the same.

Basically we are using the list of old object ids to prune any `refs/keep-around` that we use for making sure that commits referenced by comments particularly aren't pruned. This avoids the situation where comments reference objects that doesn't exist anymore, but makes clean up challenging.

@newren if you have any ideas for how to get a list of modified commits, that would be very helpful!

newren:
If you just want a list of modified commits, that's pretty easy.  Could you verify that's what you want?  @TobiX described object-id-map.old-new.txt as a complete map from all old _blob_ IDs to all new blob IDs, but in my testing it never contains any blob IDs at all.  It only ever includes commits, trees, and tag objects.  I can assume that mentioning blobs was a mistake, and the linked GitLab doc suggests gc reasons which also implies neither blobs nor tags would matter, but is it just modified commits that you need?

Anyway, what you're looking for is basically the commit-map.  At the end of filter-repo's run, it creates a .git/filter-repo/commit-map file -- and also a .git/filter-repo/ref-map file.  The former contains the mapping of how all commits were (or were not) changed, while the ref-map contains the mapping of how all local references were (or were not) changed.  I think you want the commit-map, but I mention the ref-map for two reasons: (1) to make sure you're aware of it and don't accidentally grab the wrong file, (2) if there is any chance you keep references to non-lightweight tags on the server that the user doesn't force push over then you might need some info from the ref-map, and (3) even though server-side you'd be more interested in the commit-map, I think the end-users would be more interested in the ref-map (because then they can easily see how their refs changed instead of trying to look through the massive list of all commits changed).

[...]

PR #117, opened 2020-06-21
git-filter-repo.txt: document mapping output

jamesramsay:
Useful commit and reference mappings are created on every run. These are
helpful in a number of situations, and should be documented so that
end-users and Git hosts can understand how to use the output.

The commit-map is particularly useful for Git hosts to override
retention mechanisms, like hidden refs. This allows end-users can purge
large files or sensitive data.

Follow up to #98

### DIFF
--- a/Documentation/git-filter-repo.txt
+++ b/Documentation/git-filter-repo.txt
@@ -332,6 +332,38 @@ Miscellaneous options
 --quiet::
 	Pass --quiet to other git commands called.
 
+OUTPUT
+------
+
+Every time filter-repo is run, files are created in the `.git/filter-repo/`
+directory. These files overwritten unconditionally on every run.
+
+Commit map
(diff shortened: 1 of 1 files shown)


newren:
You've got a missing word or something in that last sentence of your commit message, "This allows end-users can purge large files or sensitive data."  Or perhaps just s/can/to/?

Also, the all-zeros hash can appear in the ref-map as well as the commit-map, with the same meaning: deletion of the ref.  (For example, if instead of removing big files you are selecting just the history of one directory such as src/module-foo/, and some branches or tags are old enough that they pre-dated the introduction of the src/module-foo/ directory.)

jamesramsay:
@newren thanks for clarifying the ref-map behavior. I've fixed the commit message too.

newren:
Thanks!

PR #123, opened 2020-06-30
Added timeshift.py contrib demo 

stevef51:
Added timeshift.py which can be used to shift a range or all of a repositories commits forwards or backwards in time

### DIFF
--- a/contrib/filter-repo-demos/timeshift.py
+++ b/contrib/filter-repo-demos/timeshift.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import subprocess
+import datetime
+import sys
+try:
+  import git_filter_repo as fr
+except ImportError:
+  raise SystemExit("Error: Couldn't find git_filter_repo.py.  Did you forget to make a symlink to git-filter-repo named git_filter_repo.py or did you forget to put the latter in your PYTHONPATH?")
+
+parser = argparse.ArgumentParser(
+          description='Offset a set of or all commits by n seconds, can be used to timeshift a repository')
+parser.add_argument('--offset', type=int,
+        help=("Offset in seconds"))
+parser.add_argument('--from-commit', type=str,
+        help=("Commit to start from"))
+parser.add_argument('--to-commit', type=str,
+        help=("Commit to finish at"))
+args = parser.parse_args()
+
+print(args)
+
+shouldOffset = args.from_commit is None
+
+def offsetDate(timestampAndUtcOffset):
+  timestampAndUtcOffset = timestampAndUtcOffset.decode('utf-8')
+  split = timestampAndUtcOffset.split(' ')
+  timestamp = int(split[0])
+  timestamp = timestamp + args.offset
+  result = str(timestamp) + ' ' + split[1];
+  result = result.encode()
+  return result
+
+def fixup_commits(commit, metadata):
+  global shouldOffset
+  commitId = commit.original_id.decode('utf-8')
+  if shouldOffset == False:
+    shouldOffset = args.from_commit == commitId
+  
+  if shouldOffset == True:
+    print('Offsetting commit ' + commitId)
+    commit.author_date = offsetDate(commit.author_date)
+    commit.committer_date = offsetDate(commit.committer_date)
+
+    if not args.to_commit is None:
+      shouldOffset = args.to_commit != commitId
+
+  else:
+    print('Skipping commit ' + commitId)
+
+fr_args = fr.FilteringOptions.parse_args(['--force'])
+filter = fr.RepoFilter(fr_args, commit_callback=fixup_commits)
+filter.run()
--- a/contrib/filter-repo-demos/README.md
+++ b/contrib/filter-repo-demos/README.md
@@ -16,6 +16,7 @@ lint-history         |Run some lint command on all non-binary files in history.
 clean-ignore         |Delete files from history which match current gitignore rules.
 filter-lamely (or filter&#8209;branch&#8209;ish) |A nearly bug compatible re-implementation of filter-branch (the git testsuite passes using it instead of filter-branch), with some performance tricks to make it several times faster (though it's still glacially slow compared to filter-repo).
 bfg-ish              |A re-implementation of most of BFG Repo Cleaner, with new features and bug fixes.
+timeshift            |Timeshift a range or all of a repositories commits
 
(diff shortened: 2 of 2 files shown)


contrib/filter-repo-demos/timeshift.py
```diff
@@ -0,0 +1,55 @@
...
+  raise SystemExit("Error: Couldn't find git_filter_repo.py.  Did you forget to make a symlink to git-filter-repo named git_filter_repo.py or did you forget to put the latter in your PYTHONPATH?")
+
+parser = argparse.ArgumentParser(
+          description='Offset a set of or all commits by n seconds, can be used to timeshift a repository')
+parser.add_argument('--offset', type=int,
+        help=("Offset in seconds"))
+parser.add_argument('--from-commit', type=str,
+        help=("Commit to start from"))
+parser.add_argument('--to-commit', type=str,
+        help=("Commit to finish at"))
```
newren, line 20: The implementation of --from-commit and --to-commit here will work for linear histories and some very special non-linear ones but is in general quite broken to rely on the order of commits as output by fast-export rather than checking for actual inclusion of commits inside a history range.

contrib/filter-repo-demos/timeshift.py
```diff
@@ -0,0 +1,55 @@
...
+args = parser.parse_args()
+
+print(args)
+
+shouldOffset = args.from_commit is None
+
+def offsetDate(timestampAndUtcOffset):
+  timestampAndUtcOffset = timestampAndUtcOffset.decode('utf-8')
```
newren, line 28: Why decode into utf-8?  The values are all numeric or a space.  Seems it'd be easier to just

timestamp, timezone = timestampAndUtcOffset.split(b' ')
return b'%d %s' % (int(timestamp)+args.offset, timezone)

contrib/filter-repo-demos/timeshift.py
```diff
@@ -0,0 +1,55 @@
...
+  split = timestampAndUtcOffset.split(' ')
+  timestamp = int(split[0])
+  timestamp = timestamp + args.offset
+  result = str(timestamp) + ' ' + split[1];
+  result = result.encode()
+  return result
+
+def fixup_commits(commit, metadata):
+  global shouldOffset
+  commitId = commit.original_id.decode('utf-8')
```
newren, line 38: decoding to utf-8 is kinda slow, even when it's already known to be ascii.  It'd be easier/faster to encode args.from_commit and args.to_commit -- especially since that only has to be done once instead of one decoding per commit in the repository.  In fact, you just need to add type=os.fsencode to the parser.add_argument() call and then it'll show up as bytes that you can compare directly.

contrib/filter-repo-demos/timeshift.py
```diff
@@ -0,0 +1,55 @@
...
+  if shouldOffset == False:
+    shouldOffset = args.from_commit == commitId
```
newren, line 40: Having both an = and == on the same line makes me prefer parentheses were added around the equality expression.  I know they technically aren't needed, but it just makes it read easier to me.  Of course, see comments above about args.from_commit and args.to_commit.

contrib/filter-repo-demos/timeshift.py
```diff
@@ -0,0 +1,55 @@
...
+  if shouldOffset == True:
+    print('Offsetting commit ' + commitId)
+    commit.author_date = offsetDate(commit.author_date)
+    commit.committer_date = offsetDate(commit.committer_date)
+
+    if not args.to_commit is None:
+      shouldOffset = args.to_commit != commitId
+
+  else:
+    print('Skipping commit ' + commitId)
```
newren, line 51: Ah, so this line is where it would make sense to have commitId decoded to utf8.  It seems like this would be a warning that would be repeated an awful lot and almost seems spammy, but I can see where you're coming from with it.  I'd rather just decode the commitIds that are used with this print rather than decoding all of them, though.

contrib/filter-repo-demos/timeshift.py
```diff
@@ -0,0 +1,55 @@
...
+
+fr_args = fr.FilteringOptions.parse_args(['--force'])
```
newren, line 53: Why are you adding --force?

contrib/filter-repo-demos/timeshift.py
stevef51: Yeah, good point - hadn't thought about non linear history repos, the repo I wrote this for is small and simple with a single master branch and timeline and the tool worked perfectly timeshifting groups of commits since the dev machine at the time had incorrect date/time

contrib/filter-repo-demos/timeshift.py
stevef51: Best I could do as a complete newbie to Python :)

contrib/filter-repo-demos/timeshift.py
stevef51: Noted, wasn't aware of the os.fsencode type

contrib/filter-repo-demos/timeshift.py
stevef51: Nice pickup, and yes c = (a == b) is clearer for sure

contrib/filter-repo-demos/timeshift.py
stevef51: Yeah, did think it might be spammy, but the timeshift tool I think would be used rarely so I found it useful to check which commits were being shifted and which weren't, over a small repo (like mine) it was fine, a large repo would definitely be spammy though

contrib/filter-repo-demos/timeshift.py
stevef51: Complete newbie to Python, I think I tried passing [] but the library then complained about no parameters - noticed other samples were passing --force, so I replicated blindly :)

newren:
> I'll perhaps change it to take --from-date and --to-date instead of commit id's, not something that I'll get around to anytime soon though, current version did what I needed it to and thought it might be helpful to someone else hence the pull-request :)

Totally understand.  Thanks for taking the time to contribute; others can search for it and perhaps find this, and if their needs are simple and won't get caught up by the edge cases I mentioned as problematic, then it may help them.  I also understand that not having the time to make it robust for all cases, so I'll go ahead and close out for now and if you want to revisit in the future we can do so.  Thanks again for contributing.  :-)

PR #127, opened 2020-07-08
.gitignore: ignore the generated documentation

tarsius:
There's not much else I could say here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
+/Documentation/html/
+/Documentation/man1/
 /t/test-results
 /t/trash directory*


newren:
Sorry for the long delay in responding...

I'm curious about the motivation here.  I never generate the documentation within git-filter-repo and don't even know how to.  Is it easy?  How are you doing it?  Personally, I just cheat and symlink git-filter-repo.txt into git.git sources and ran make from its Documentation/ subdirectory.

Other than that, the patch and commit message look good but I am curious more about why or how this affects you (and maybe if you're doing something better I should copy...)

tarsius:
> Sorry for the long delay in responding...

Well there's this other issue where I am taking an even longer time, so...

> I'm curious about the motivation here. I never generate the documentation within git-filter-repo and don't even know how to. Is it easy? How are you doing it? Personally, I just cheat and symlink git-filter-repo.txt into git.git sources and ran make from its Documentation/ subdirectory.

I'm just using your `update-docs` make target and I have a symlink in `~/man/man1` to actually be able to easily read the manual page. It's a hack on top of your hack I am afraid. Unfortunately I haven't figured out anything that could help us to generate/install the documentation more gracefully.

newren:
Well, it certainly doesn't hurt anything, and if it makes it easier for you then it seems fine.  I went ahead and merged it; thanks!

tarsius:
Cheers!

Issue #128, opened 2020-07-09
Three undefined names in Python code


cclauss:
[flake8](http://flake8.pycqa.org) testing of https://github.com/newren/git-filter-repo on Python 3.8.3

$ __flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics__
```
./git_filter_repo.py:2804:14: F821 undefined name 'callback'
      return callback #namespace['callback']
             ^
./git_filter_repo.py:3519:51: F821 undefined name 'branch'
                         (decode(marks_basename), branch))
                                                  ^
./t/t9391/erroneous.py:12:44: F821 undefined name 'progress'
  print("Decipher this: "+''.join(reversed(progress.message)))
                                           ^
3     F821 undefined name 'callback'
3
```
https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead, these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python, they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python, a __NameError__ is raised which will halt/crash the script on the user.

newren:
I apologize for the lengthy delay in getting back to your report, especially given how high quality it is.  Thanks for taking the time to report these issues and for the pointer to flake8; it seems very cool.  Of the three cases reported:

* the last was a case meant to have an error, but somewhere else that executes earlier.  The highlighted line of code just didn't make sense and was clearly a copy-and-paste from another testcase.  While the line doesn't get executed because of an earlier error, this just makes the demos harder to understand.  I just rewrote that callback.
[...]

PR #130, opened 2020-07-10
Undefined name: branch --> full_branch

cclauss:
As discussed in #128

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -3516,7 +3516,7 @@ class RepoFilter(object):
         contents = subproc.check_output(cmd)
       except subprocess.CalledProcessError as e: # pragma: no cover
         raise SystemExit(_("Failed loading %s from %s") %
-                         (decode(marks_basename), branch))
+                         (decode(marks_basename), full_branch))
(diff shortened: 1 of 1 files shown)


newren:
Sorry for the delay in responding.  Thanks for the awesome bug report and the patch.  The diff looks good, but the commit message needs some touch-ups:
  * As per git.git guidelines, use an "area: " prefix and follow with imperative tense
  * As per git.git guidelines, please provide a Signed-off-by on your commit message (https://github.com/git/git/blob/master/Documentation/SubmittingPatches#L300)

So the commit message might look something like: """
filter-repo: fix typo to correct undefined variable name

Signed-off-by: $YOUR_NAME <$YOUR_EMAIL>
"""

newren:
I'm assuming that your deletion of the branch means you'd rather not take the time to fix it up, probably related to how slow I was to respond.  My apologies.  I fixed up both the problems you reported in commit e4960a53f8 (Fix undefined variable names, 2020-07-27), and recorded you as the reporter of the issues.

Thanks for the reports!

PR #135, opened 2020-07-31
Handle encoding of unicode/unicode big endian/utf8

LarrysGIT:
When the parameter --paths-from-file point to a non-ASCII text file, the first line start with extra bytes indicates the file's encoding. The following 3 types are handled.

```
        # Unicode
        line = line.lstrip(b'\xff\xfe1\x00')
        # Unicode big endian
        line = line.lstrip(b'\xfe\xff\x001')
        # utf8
        line = line.lstrip(b'\xef\xbb\xbf1')
```

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -2118,6 +2118,13 @@ EXAMPLES
       for line in f:
         line = line.rstrip(b'\r\n')
 
+        # Unicode
+        line = line.lstrip(b'\xff\xfe1\x00')
+        # Unicode big endian
+        line = line.lstrip(b'\xfe\xff\x001')
+        # utf8
+        line = line.lstrip(b'\xef\xbb\xbf1')
+
         # Skip blank lines
         if not line:
           continue


LarrysGIT:
> Thanks for sending in the change; sorry for taking a while to review it.
> 
> As mentioned in the other comment, the commit message needs a lot more explanation if the code is actually correct, though I suspect there are problems. I'd also need the commit to be done not by a sample author name but a real name, and you'd need to provide your Signed-off-by as per the contribution guidelines.

Appreciate you didn't approve the pull request, I didn't put many tests and thoughts on this fix. I just checked it again and find the fix is invalid.
The bytes to trim is incorrect, correct marker bytes are following
```
# unicode -> b'\xff\xfe'
# unicodebigendian -> b'\xfe\xff'
# utf8 -> b'\xfe\xff'
```
However, I have re-considered this fix, it will require many code changes and add extra dependencies to handle all file encodings, which I think it is not worth doing.

PR #147, opened 2020-08-25
lint-history: note that filename is absolute

ruv:
Describe `filename` argument for `--relevant` option.

### DIFF
--- a/contrib/filter-repo-demos/lint-history
+++ b/contrib/filter-repo-demos/lint-history
@@ -59,6 +59,9 @@ example_text = '''CALLBACK
         def is_relevant(filename):
             BODY
 
+    Where filename is a fully qualified file name. The file is located in a temporary directory
+    that can be specified via TMPDIR environment variable.
+
     Thus, to only run on files with a ".txt" extension you would run
         lint-history --relevant 'return filename.endswith(b".txt")' ...
 


contrib/filter-repo-demos/lint-history
```diff
@@ -59,6 +59,9 @@ example_text = '''CALLBACK
         def is_relevant(filename):
             BODY
 
+    Where filename is a fully qualified file name. The file is located in a temporary directory
+    that can be specified via TMPDIR environment variable.
```
newren, line 63: I think this will be somewhat confusing to users; they'll see it and expect "/Users/myusername/projects/clone-of-repo/path/to/code.c" instead of "path/to/code.c".  So we should probably instead mention that it is a path that is relative to the toplevel of the repository (as are all paths in filter-repo).

The first sentence and second sentences also seem jarring next to each other without more explanation.  The second sentence is unrelated to CALLBACKS and is true whether or not users specify the --relevant flag, so we should put that other sentence somewhere else; perhaps making a new section after "EXAMPLES" named "INTERNALS" or something.  And perhaps tweak the wording to make it a bit more clear, e.g. "Historic versions of files being linted will be written to a temporary directory before running the linting program; the location of this temporary directory can be controlled via the TMPDIR environment variable as per https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp."

newren (changes requested):
Hi, thanks for improving the documentation!  Sorry for taking a while to respond.  I think there are a few tweaks we'd want to make; I commented on the lines in particular.  Plus there's a couple changes needed to the commit itself:
  * Can you add a real email address?  I suspect "2678296+ruv@users.noreply.github.com" is something auto-generated by github when doing inline editing.
  * Can you add your signed-off-by to the commit ("git commit --amend -s" ought to do it), as noted in the contribution guidelines?

dscho:
@newren just a quick note from one maintainer to another: you might have missed that most people opening PRs allow maintainers to edit those PRs by (force-)pushing to their branches in their forks. You might want to do that here, to avoid the PR becoming too stale.

contrib/filter-repo-demos/lint-history
ruv: IIRC, in my test the file name was full (absolute), i.e. from the root directory, and the path didn't repeat the hierarchy from the project (i.e., `path/to/` in your example). Perhaps, it should be checked again.
Knowing path in the project is a good thing, since in some cases a transformations should be applied to some specific files in the project only.

Concerning TMPDIR — yes, agreed.

ruv:
@newren @dscho  thanks for review and comments. Yes, actually you may edit my PR. Or just reject this PR and make changes by you own. It was just a quick online editing of the file in GitHub, to show my idea. At the moment I don't have a time to make this PR better, sorry.

newren:
Fixed in commits ccc37d3423ac (lint-history: explain filename paths, 2021-06-05) and e5d8938d48b3 (lint-history: explain how TMPDIR can be used, 2021-06-05)

PR #149, opened 2020-08-26
docs(collection): update docs around collecting and installing doc files

### LINKED ISSUE #140: No git-repo-filter.html file in root
In the INSTALL.md instruction there is a reference to git-filter-repo.html (and to git-filter-repo.1) both are **not found** in the (root of) the repository.

joshsleeper:
simple update to clarify how to obtain the doc files

closes #140 (which is where I learned what I needed to do~)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -51,6 +51,9 @@ filter-repo only consists of a few files that need to be installed:
     displaying the manpage, when help.format is "man" (the default on Linux
     and Mac).
 
+    `make snag_docs` must be run to retrieve this file, depositing it
+    at `Documentation/man1/git-filter-repo.1`
+    
     This can be installed in the directory pointed to by `$(git
     --man-path)/man1/`, or placed anywhere in $MANDIR/man1/ where $MANDIR
     is some entry from $MANPATH.
@@ -64,6 +67,9 @@ filter-repo only consists of a few files that need to be installed:
     displaying the html version of the help, when help.format is set to
     "html" (the default on Windows).
 
+    `make snag_docs` must be run to retrieve this file, depositing it
+    at `Documentation/html/git-filter-repo.html`
+
     This can be installed in the directory pointed to by `git --html-path`.
 
     Note that `git filter-repo -h` will show a more limited built-in set of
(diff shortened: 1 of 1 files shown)


roccomoretti:
This will not work for people who have downloaded one of the released versions from https://github.com/newren/git-filter-repo/releases, as `make snag_docs` will not work if you're not in a git clone.

joshsleeper:
ah, yea good catch. it looks like in the versioned release artifact `make snag_docs` has already been executed so the files exist.

what do you think about me modifying the language to say that `make snag_docs` is required only if you've cloned the repo locally?

INSTALL.md
```diff
@@ -51,6 +51,9 @@ filter-repo only consists of a few files that need to be installed:
     displaying the manpage, when help.format is "man" (the default on Linux
     and Mac).
 
+    If you've cloned the repo directly `make snag_docs` must be run to
```
newren, line 54: "must be" isn't true; "can be" would be better here.

INSTALL.md
```diff
@@ -64,6 +67,9 @@ filter-repo only consists of a few files that need to be installed:
     displaying the html version of the help, when help.format is set to
     "html" (the default on Windows).
 
+    If you've cloned the repo directly `make snag_docs` must be run to
```
newren, line 70: same -- "can" instead of "must".

(Folks can build the docs manually, even if not documented how to do so.  They could also check them out from the other branch or download them from online, or just ignore these documentation files.)

INSTALL.md
```diff
@@ -73,10 +79,14 @@ filter-repo only consists of a few files that need to be installed:
 So, installation might look something like
 ```
 cp -a git-filter-repo $(git --exec-path)
-cp -a git-filter-repo.1 $(git --man-path)/man1
-cp -a git-filter-repo.html $(git --html-path)
 ln -s $(git --exec-path)/git-filter-repo \
     $(python -c "import site; print(site.getsitepackages()[-1])")/git_filter_repo.py
+
+# If you've cloned the repo directly you need to make the doc files
```
newren, line 85: Perhaps change this to "If you've cloned the repo directly, grab the documentation files via".

newren (changes requested):
Sorry for the long delay.  Overall, I like the changes with a few small wording tweaks.  Thanks for sending this!

A couple other small changes are needed:
  * Please add a Signed-off-by tag to the commit message as per contributing guidelines
  * Please use a real email address; I'm guessing GitHub put @users.noreply.github.com in there from inline editing or something.
  * Please remove the whitespace error(s) (visible from 'git log -p --check')
  * Please squash the two commits together to make the history clean.

newren:
Can't merge without the Signed-off-by, so I'll go ahead and close. Feel free to reopen if you're willing to change the commit message to include your signed-off-by.

joshsleeper:
junk, this totally fell off my radar until you closed it earlier, sorry!

I've re-created the PR with most of the desired changes as #219, sorry for the hassle

PR #151, opened 2020-09-01
Fix typo

tomtastic:


newren (changes requested):
Sorry for the long delay; thanks for catching this and sending in a patch!

Can you make two small tweaks:
  * Make the commit message be "filter-repo: fix typo" as per contributing guidelines
  * Add a Signed-off-by line to the commit message ('git commit --amend -s' will do it), as per contributing guidelines

If you make those two changes, I'll merge it.  Thanks much for the contribution!

PR #154, opened 2020-09-08
filter-repo: only set author from committer if author_email not set

mwilck:
Some commits may have a valid author_email, but no valid author_name.
Setting the author data from the committer is wrong in this case.

Example: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c6295cdf656de63d6d1123def71daba6cd91939c

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1189,13 +1189,14 @@ class FastExportParser(object):
       original_id = self._parse_original_id();
 
     author_name = None
+    author_email = None
     if self._currentline.startswith(b'author'):
       (author_name, author_email, author_date) = self._parse_user(b'author')
 
     (committer_name, committer_email, committer_date) = \
       self._parse_user(b'committer')
 
-    if not author_name:
+    if not author_name and not author_email:
       (author_name, author_email, author_date) = \
         (committer_name, committer_email, committer_date)
 
--- a/t/t9390/unusual
+++ b/t/t9390/unusual
@@ -32,4 +32,13 @@ tagger little.john <> 1535229618 -0700
 data 4
 v1.0
 
+commit refs/heads/other
+mark :3
+original-oid 0000000000000000000000000000000000000004
+author  <me@little.net> 1535228578 +0500
+committer Srinivasa Ramanujan <fellow@cambridge.org> 1535228580 +0500
+data 8
+Initial
+M 100644 :1 greeting
+
 done
--- a/t/t9390/unusual-filtered
+++ b/t/t9390/unusual-filtered
@@ -3,9 +3,18 @@ blob
 mark :1
 data 5
 hello
+reset refs/heads/other
+commit refs/heads/other
+mark :2
+author  <me@little.net> 1535228578 +0500
+committer Srinivasa Ramanujan <fellow@cambridge.org> 1535228580 +0500
+data 8
+Initial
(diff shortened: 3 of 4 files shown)


mwilck:
The first version I pushed was wrong, it would crash if no author data was given at all, like in the "unusual" test case. Now I'm adding a test case for the situation i'm trying to solve. 

8dcaeed is unrelated to this issue; it is a trivial fix for the fact that tests were failing in my environment because the system-installed git-filter-repo was run rather than that from my repository.

mwilck:
Note: I would have liked to keep the test case changes smaller. But for some reason I didn't grok, if I added my new entry at the end of `unusual`, it would still always be ordered before the existing ones in the output.

newren:
Thanks, this is great!  And sorry for taking so long to respond.  Overall this looks really good; I like the code changes.  There's just two small problems with the test changes:

Unfortunately, there's also a test in t9391 that relies on t9390/unusual and digs into some internals, and the extra objects break some checks there, failing that test.  It'd be nicer if we could just create a new test, or at least modify one of the ones other than "unusual".

Commit 8dcaeed ("t9390: run git-filter-repo from TEST_DIRECTORY"), which you marked as unrelated isn't what we want. "git filter-repo" is used throughout the testsuite, so we'd need to change all of them if we wanted to go that route, but since the documentation uses the form with a space I'd like to make sure we test it.  Instead, we could add some documentation about the needed setup for running the tests, or make some kind of tweak to the test harness to set GIT_EXEC_PATH to pick up the right git-filter-repo.  However, since as you mention that's unrelated, could you just drop that patch from this series and if you want put it in another issue or pull request where we can discuss that separately?

Okay, there might be a third minor issue -- it'd also be nice of the code and test changes were in the same patch, so that the testsuite remained passing with every commit.  Or, if you want to keep the testsuite change in a separate commit, then mark it as test_expect_failure, and in the patch that makes the code change it will flip the new test from test_expect_failure to test_expect_success.

Thanks for working on this!

newren (changes requested):
(Just adding another comment so I can mark as changes-requested.)

newren:
> Note: I would have liked to keep the test case changes smaller. But for some reason I didn't grok, if I added my new entry at the end of `unusual`, it would still always be ordered before the existing ones in the output.

Oh, I forgot to comment on this earlier.  I'm pretty sure this is related to the fact that fast-export does topological sorting and then it picks which branch from there to go down.  That often means it'll take branches with less history in them and cover them first.  Since the test uses fast-export and compares to its output, that means we're subject to its behavior.  Anyway, not a big deal and I would have been fine with it, but the fact that other ugly internals testing was done on that testcase in t9391 just meant we really needed to do the testing with some other test or a new one.  You just got unlucky with the one you picked.

mwilck:
@newren, thanks for your response. This is just to notify you that I saw it, and that it makes sense to me. I'm just not sure when I'll be able to get around to work on it.
For now, I've dropped 8dcaeed, and squashed the code change and the test code addition into one commit.

newren:
I replaced the testcase changes with a new dedicated test, and then merged it.  Result can be seen as commit 282f8ddb9b (filter-repo: only set author from committer if author email not set, 2020-09-09), so I'll close this one out.

Thanks for the contribution!

PR #161, opened 2020-10-07
filter-lamely: fix NameError because of forgotten fr module prefix

RagingCactus:
I have succesfully used `git-filter-repo` for cleanup after migrating a lot of repos from SVN to Git and subsequent consolidation of many small repos into bigger combined repos. Thanks again for maintaining this useful tool!

I even promise I did most of that work using proper `git-filter-repo` commands! However, I recently had to reach for `filter-lamely` again for line ending conversion and noticed that it breaks when the repo has annotated tags because of a missing module prefix. This PR should fix that.

### DIFF
--- a/contrib/filter-repo-demos/filter-lamely
+++ b/contrib/filter-repo-demos/filter-lamely
@@ -575,7 +575,7 @@ class UserInterfaceNightmare:
       return
 
     tag.skip()
-    reset = Reset(tag.ref, tag.from_ref)
+    reset = fr.Reset(tag.ref, tag.from_ref)
     self.filter.insert(reset, direct_insertion = False)
 
   def muck_stuff_up(self):


newren:
Sweet, thanks for sending the fix along; I merged it.

As a side note: If you're using filter-lamely so that you can shell out to other commands (like dos2unix or code linters or whatever), note that [lint-history](../blob/main/contrib/filter-repo-demos/lint-history) may be a better option (avoids a lot of the safety problems, updates hashes in commit messages, probably has better performance, might be easier to use, etc.).  However, using whatever works for you is always fine, and it's even better to send along fixes to make it better for everyone else.  Thanks for doing that.

RagingCactus:
In this particular instance I mainly use filter-lamely to leverage git's own line ending conversion in a tree filter as I have some experience using it and know that it does what I need. It may very well be possible to do this with proper filter-repo callbacks, but for my particular use case I can get away with the drawbacks (tags will be discarded and merged into existing branches, single-digit amount of commit hashes in commit messages anyway, history is not extremely long, ...)
If you have a quick reference handy for this use case I'd appreciate it.

In any case, thanks for merging and serving as a maintainer for this useful tool! filter-repo definitely saved the day for the migration from SVN to Git at my day job

PR #162, opened 2020-10-08
filter-repo: Support GIT_COMMIT env-var for filter callbacks

nipunn1313:
GIT_COMMIT is one of a variety of environment variables supported
by git-filter-branch documented in
https://git-scm.com/docs/git-filter-branch#_filters

Filter callbacks may now access this env var

Signed-off-by: Nipunn Koorapati <nipunn@dropbox.com>

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -3357,6 +3357,11 @@ class RepoFilter(object):
     commit.file_changes = [v for k,v in sorted(new_file_changes.items())]
 
   def _tweak_commit(self, commit, aux_info):
+    # Callbacks are provided environment variables related to commit
+    # similarly to filter-branch
+    # https://git-scm.com/docs/git-filter-branch#_filters
+    os.environ["GIT_COMMIT"] = commit.original_id.decode('utf-8')
+
     # Change the commit message according to callback
     if not self._args.preserve_commit_hashes:
       commit.message = self._hash_re.sub(self._translate_commit_hash,
--- a/t/t9392-python-callback.sh
+++ b/t/t9392-python-callback.sh
@@ -64,9 +64,9 @@ test_expect_success '--message-callback' '
 	setup message-callback &&
 	(
 		cd message-callback &&
-		git filter-repo --message-callback "return b\"TLDR: \"+message[0:5]" &&
+		git filter-repo --message-callback "return b\"TLDR: \"+message[0:5]+b\" \"+os.environ[\"GIT_COMMIT\"].encode(\"utf-8\")" &&
 		git log --format=%s >log-messages &&
-		grep TLDR:...... log-messages >modified-messages &&
+		grep "TLDR:...... .\{40\}" log-messages >modified-messages &&
 		test_line_count = 6 modified-messages
 	)
 '


nipunn1313:
Hi - I'm new to contributing to git / the git family of projects. I did my best to follow along with the guidelines, but bear with me if I missed something. Thanks!

I put this together because as I migrated an old script from git filter-branch to git filter-repo, I needed access to the GIT_COMMIT commit_id from the callback.

`git-filter-repo` is a great project! 100s of times faster for what we needed it for. Thanks for putting time into it.

t/t9392-python-callback.sh
```diff
@@ -64,9 +64,9 @@ test_expect_success '--message-callback' '
 	setup message-callback &&
 	(
 		cd message-callback &&
-		git filter-repo --message-callback "return b\"TLDR: \"+message[0:5]" &&
+		git filter-repo --message-callback "return b\"TLDR: \"+message[0:5]+b\" \"+os.environ[\"GIT_COMMIT\"].encode(\"utf-8\")" &&
```
newren, line 67: Without the change to _tweak_commit() that you made, you could just use a commit callback in order to have access to all information about the commit, and make the same kind of tweak to the commit message:

git filter-repo --commit-callback "commit.message = b\"TLDR: \"+commit.message[0:5]+b\" \"+commit.original_id"

git-filter-repo
```diff
@@ -3357,6 +3357,11 @@ class RepoFilter(object):
     commit.file_changes = [v for k,v in sorted(new_file_changes.items())]
 
   def _tweak_commit(self, commit, aux_info):
+    # Callbacks are provided environment variables related to commit
+    # similarly to filter-branch
+    # https://git-scm.com/docs/git-filter-branch#_filters
+    os.environ["GIT_COMMIT"] = commit.original_id.decode('utf-8')
```
newren, line 3363: Environment variables are a way of passing information between parent and child processes. Callbacks run in the same process, so using environment variables to pass information around seems like a wasteful way to pass the information.  Especially since a commit callback can just access commit.original_id directly.

newren (commented):
Not sure I understand why this would be wanted when you can just use commit.original_id.  Perhaps you were unaware that --message-callback is just a convenient shorthand for a subset of what you can access with --commit-callback, and that you can modify the commit message (and much more) with the later, while also accessing the commit.original_id value directly?

t/t9392-python-callback.sh
nipunn1313: Aha! Since I was only intending on changing the message, I had sorta tunneled on the `--message-callback` section of the docs, failing to see the `--commit-callback` - yep this makes sense!

I wanted the message callback to have access to the old commit id - so that we could leave a trace of the old commit ID in the rewritten message. It appears that `-message-callback` isn't good for this, but `--commit-callback` is!

I may be able to propose a piece of documentation that may make this more obvious (to me), but for now I'll close this!

nipunn1313:
Thanks for the review and suggestion!

nipunn1313:
Documentation proposal: #164

PR #164, opened 2020-10-09
filter-repo doc: Document --name-callback as shorthand

nipunn1313:
Make it easier to observe how --name-callback relates
to --commit-callback

Signed-off-by: Nipunn Koorapati <nipunn@dropbox.com>

### DIFF
--- a/Documentation/git-filter-repo.txt
+++ b/Documentation/git-filter-repo.txt
@@ -1039,6 +1039,15 @@ def name_callback(name):
   return name.replace(b"Wiliam", b"William")
 --------------------------------------------------
 
+The above `--name-callback` incanatation is shorthand for this raw `--commit-callback`
+
+--------------------------------------------------
+git-filter-repo --commit-callback '
+	commit.author_name = commit.author_name.replace(b"Wiliam", b"William")
+	commit.committer_name = commit.committer_name.replace(b"Wiliam", b"William")
+'
+--------------------------------------------------
+
 The email callback is quite similar:
 
 --------------------------------------------------


Documentation/git-filter-repo.txt
```diff
@@ -1039,6 +1039,15 @@ def name_callback(name):
   return name.replace(b"Wiliam", b"William")
 --------------------------------------------------
 
+The above `--name-callback` incanatation is shorthand for this raw `--commit-callback`
```
nipunn1313, line 1042: One think I noticed is that this message happens before `--commit-callback` is explained below. I think this is ok, since it's referenced above initially, so you'll know where to look.

Documentation/git-filter-repo.txt
```diff
@@ -1039,6 +1039,15 @@ def name_callback(name):
...
+
+--------------------------------------------------
+git-filter-repo --commit-callback '
+	commit.author_name = commit.author_name.replace(b"Wiliam", b"William")
+	commit.committer_name = commit.committer_name.replace(b"Wiliam", b"William")
+'
+--------------------------------------------------
```
newren, line 1049: This is not correct; as stated in the manual: "The other callbacks operate on a small piece of the raw objects or operate on pieces across multiple types of raw object (e.g. author names and committer names and tagger names across commits and tags..."  Thus, the name-callback operations on author name and commiter name of commits AND tagger name of tags.

The message callback is similar -- it operations on commit messages of commits AND tagger messages of tags.  If someone wants to pick off just one type of name or just one type of message, they should use the commit or tag callbacks.

newren (changes requested):

Documentation/git-filter-repo.txt
nipunn1313: Yes that makes sense - thanks for the clarification.
I think I'll let this one go then.

PR #167, opened 2020-10-21
Add a GitHub workflow for continuous testing

dscho:
The idea is to fix the tests so that they pass on Windows, and then have a GitHub workflow that runs on every push and Pull Request to ensure that no regressions are introduced on the main three platforms: Windows, macOS and Linux.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+*.sh eol=lf
+*.py eol=lf
+/git-filter-repo eol=lf
+/contrib/filter-repo-demos/[a-z]* eol=lf
+/t/t9*/* eol=lf
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,49 @@
+name: Run tests
+
+on: [push, pull_request]
+
+jobs:
+  run-tests:
+    strategy:
+      matrix:
+        os: [ 'windows', 'ubuntu', 'macos' ]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.x
+    - name: test
+      shell: bash
+      run: |
+        # setup-python puts `python` into the `PATH`, not `python3`, yet
+        # `git-filter-repo` expects `python3` in the `PATH`. Let's add
+        # a shim.
+        printf '#!/bin/sh\n\nexec python "$@"\n' >python3 &&
+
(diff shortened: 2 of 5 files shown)


dscho:
@newren I plan on working on this a little bit more. Please note that this PR takes the unrefined work I presented in https://lore.kernel.org/git/nycvar.QRO.7.76.6.2010211630010.56@tvgsbejvaqbjf.bet/ and refines it _a little_ more.

dscho:
TODO: use https://github.com/marketplace/actions/setup-python instead of downloading the full-fledged Git for Windows SDK (which takes ~3m20, quite a bit longer than actually running the tests).

dscho:
@newren while I have your attention... t9390.43 seems to fail for me all the time, even in WSL (i.e. Linux):

```
[...]
++ test_line_count = 5 ../filenames
++ test 3 '!=' 3
+++ wc -l
++ test 6 = 5
++ echo 'test_line_count: line count for ../filenames != 5'
test_line_count: line count for ../filenames != 5
++ cat ../filenames

numbers/medium.num
numbers/small.num
sequence/know
whatever
words/know
++ return 1
error: last command exited with $?=1
not ok 43 - --strip-blobs-with-ids
```

Is this test passing for you?

dscho:
> Is this test passing for you?

Oh wait, it passes for me on Linux with v2.29.0. Let me investigate.

dscho:
Ahhh! 2ca6e362bdbe7e938092e192ee68fde823c33c48 is the "culprit" ;-)

newren:
Ah, that '&&' was masking a bug in the testcase.  Not sure how it got in there, but digging through it, the correct count is 6.  The failure is in part due to the file named 'mercurial'; it was introduced in an "evil" merge and didn't exist in either parent.  But my method for checking which files existed was kinda lame.

In addition to adding the '&&' and changing the expected count to 6, I should probably add a
```
		test_must_fail git rev-parse HEAD~4:mercurial &&
```
line to the test to ensure the mercurial file was ripped out.

dscho:
Right. I actually managed to fix that, and pinpoint it to the commit that introduced the test case: b6a35f8

dscho:
There you go.

dscho:
@newren would you kindly review this thing? I know, you don't like GitHub's PR UI because it does not let you comment easily on commit messages. Hopefully there is not a lot you need to comment about, though.

newren:
Sweet, thanks @dscho; this is awesome!

newren:
...and thanks for writing such great commit messages, as always.  I learned a couple things about MSYS2 and Windows; I think one of those might be relevant to one of the issues that was filed if I can find it...

PR #206, opened 2021-02-18
Fix bug in --path-rename argument without colon

lassik:
When `git filter-repo --path-rename foo` is given, and `foo` does not have a colon, an exception with the following stack trace crashes the program.

```
  File "/usr/local/bin/git-filter-repo", line 1626, in __call__
    if values[0] and values[1] and not (
IndexError: list index out of range
```

This commit fixes the bug, and also checks against the user giving more than one colon in the argument (that would probably indicate some kind of mistake).

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1622,7 +1622,10 @@ class FilteringOptions(object):
       if suffix.startswith('rename'):
         mod_type = 'rename'
         match_type = option_string[len('--path-rename-'):] or 'match'
-        values = values.split(b':', 1)
+        values = values.split(b':')
+        if len(values) != 2:
+          raise SystemExit(_("Error: --path-rename expects one colon in its"
+                             " argument: <old_name:new_name>."))
         if values[0] and values[1] and not (
            values[0].endswith(b'/') == values[1].endswith(b'/')):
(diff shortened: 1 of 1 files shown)


lassik:
The Windows test failed in CI. Is there some situation where a drive letter like `C:` needs to go into the argument?

newren:
Hi, and thanks for the contribution!  My apologies for being slow to get back to you.

I suspect the windows GitHub Actions was just being flaky, to be honest.  Had one other report from the same time period where simply re-running made it pass.

The code change looks obviously good.  The commit message needs both a Signed-off-by line (as per https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches#n303) and a proper description instead of just a stack trace.  Perhaps something like the following in the body of the commit mesage:

The --path-rename flag expected an argument with a colon character (':') in it, which it assumed without checking.  Add some error checking with a nicer message for users than a traceback if this assumption is violated.

Signed-off-by: Lassi Kortela <whatever-your@email.address.is>

If you can make such changes to your commit message, I'd be happy to merge.  Thanks again!

lassik:
No prob. I amended the commit message as you suggest. Though the code is unchanged, the CI test now passes.

newren:
Thanks!

PR #208, opened 2021-02-19
Remove redundant end indices on list slices

ugultopu:


ugultopu:
Sorry, I missed that one keeps it as a list, while the other just converts it to its type.

ugultopu:
Actually just noticed that they are strings, so slicing one element and indexing an element is the same thing.

However, list slice returns an empty string when indices are out of range, while indexing throws and error when index is not present in the string. Hence, the tests fail when we use list index instead of list slice.

PR #216, opened 2021-03-07
 INSTALL.md: add installation via Homebrew

nurdabolatov:


newren:
Hey, thanks for the pull request.  In this case, though, installation via brew was already covered -- see https://github.com/newren/git-filter-repo/blob/main/INSTALL.md#installation-via-package-manager.

PR #222, opened 2021-03-11
Debug GitHub workflows

newren:


dscho:
@newren do you know about https://github.com/marketplace/actions/debugging-with-tmate? It makes debugging substantially easier, I find.

newren:
@dscho: ooh, sweet, thanks for the link.

PR #224, opened 2021-03-16
Improve installation documentation for Windows users

ViliusS:
Since Windows do allow to install Python more easily these days and takes care of most PATH stuff, use recommendations from Microsoft.

The only hiccup is that Python from the store cannot parse get-filter-repo SHEBANG value when the script is installed via pip. That value is very long by default. Changing it to #!\python.exe solves the issue.

Tested under clean installation of Windows 10 Enterprise version 20H2.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,8 +19,33 @@ these versions.
 
 # Notes for Windows Users
 
-It seems that the default python on Windows is not fully functional,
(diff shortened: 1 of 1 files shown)


ViliusS:
Sorry commited from the wrong account. Will make another pull request shortly.

PR #217, opened 2021-03-09
Use setup.py entry_points for installation

qwerbzuio:
This should make the installation via pip more robust.

On Windows the usage of entry_points will install a wrapper executable
for the script that chooses the proper python executable. This
essentially makes the script run correctly when called via `git
filter-repo` (direct execution via `git-filter-repo` was already fine
before).

This fixes #144 where git (on Windows) chooses a different python
executable than the one indicated by the installation via `pip{x,3}
install`.

----

I have no overview of all the installation methods and how this PR might
influence them. Also, I obviously couldn't verify the entire deployment process
to PyPI, so this still needs to be tested. I could only check that the created
wheel installs properly on Windows (with some local modifications to the
Makefile, as it seems to be designed for usage on Linux only).

Issue #193, opened 2021-01-21
Can't get it to work on Windows


Slion:
All I get is `error: cannot spawn git-filter-repo: Exec format error`.
Used Python as provided from Microsoft Store.
Installed git-filter-repo using pip.
Copied script to git folder.
Try run it.
Get error mentioned above.

Tried to change the shebang line, no joy.

Slion:
Shebang: `#!crap`
Result: `error: cannot spawn git-filter-repo: Exec format error`

Shebang: `#!C:\Users\username\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\python.exe`
Result: `error: cannot spawn git-filter-repo: Exec format error`

Shebang: `#!/usr/bin/env python`
Result: `/usr/bin/env: 'python': Permission denied`

Slion:
Shebang: `#!crap`
Result: `error: cannot spawn git-filter-repo: Exec format error`

Shebang: `#!C:\Users\username\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\python.exe`
Result: `error: cannot spawn git-filter-repo: Exec format error`

Shebang: `#!/usr/bin/env python`
Result: `/usr/bin/env: 'python': Permission denied`

newren:
Did you read https://github.com/newren/git-filter-repo/blob/main/INSTALL.md (linked to from the main page) and particularly the NOTES for windows users?  Did you try things mentioned in the three links from there?

Slion:
> Did you read https://github.com/newren/git-filter-repo/blob/main/INSTALL.md (linked to from the main page) and particularly the NOTES for windows users? Did you try things mentioned in the three links from there?

I did read through those yes. Did not help though. I did a rebase instead 😁

devkoa:
Hello there, @Slion you are not right, it works in Windows. In my case i have hundred of git repositories, and i develop and test in Windows system. And i have a quite complex setup. I run python code, i generate a shell script and the shell script calls the git-filter-repo, and all this works well in Windows. So i dont see any issue here. 
@newren Thanks, great work.

SteveBenz:
I haven't figured this all the way out, but in Python 3.9.0, which is the one I downloaded from python.org (not the windows store version), it seems that the default instructions of replacing:

#!/usr/bin/env python3

[...]

Issue #189, opened 2021-01-08
An installation guide that makes sense for windows users


PaulVrugt:
I've been trying to use git-filter-repo on windows, but hints on how to install it on windows are scattered around the internet. It would be very nice to have a simple step by step guide on how to install this tool on windows.

I've managed to use `pip` to use python (which is not a really common tool to use for windows users) to install git-filter-repo on my machine, but I've been unable to get git to recognize the `filter-repo` command

newren:
Did you look through the "NOTES for windows users" under https://github.com/newren/git-filter-repo/blob/main/INSTALL.md?

If so, the best I can do is ask for volunteers to write such an installation guide, as I have no windows machines myself, but the three links in that section appear to be cases where others have done so, documenting the problems and their fixes.

newren:
[I updated the INSTALL.md file](https://github.com/newren/git-filter-repo/blob/main/INSTALL.md#simple-installation) to point out that folks can just run `python3 git-filter-repo` instead of trying to get `git filter-repo` to work.  Don't know why Windows makes it so difficult to make the short form magic work, but it's really not necessary.

Issue #144, opened 2020-08-22
"SyntaxError: invalid syntax" after installing 2.28 with PIP on Windows 10


chriskorosu:
C:\Users\cms>pip install git-filter-repo
You are using pip version 6.0.8, however version 20.2.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting git-filter-repo
  Downloading https://files.pythonhosted.org/packages/c7/a3/f5a470387c6b9c6d560b74d6cee21d56d595c970a439e5e702b595e520a8/git_filter_repo-2.28.0-py2.py3-none-any.whl (97kB)
    100% |################################| 98kB 2.5MB/s
Installing collected packages: git-filter-repo

Successfully installed git-filter-repo-2.28.0

C:\Users\cms>git filter-repo -h
  File "C:/Program Files/Git/mingw64/libexec/git-core\git-filter-repo", line 3236
    **extra_items}
     ^
SyntaxError: invalid syntax

C:\Users\cms>

qwerbzuio:
I get this very error with python 3.7.9 or python 3.8.5 and git-filter-repo 2.29.0. Output is:
```
$ git filter-repo
  File "C:\Users\bmz3\.local\bin\git-filter-repo", line 3239
    **extra_items}
     ^
SyntaxError: invalid syntax
```

OS is also Windows 10. I installed git-filter-repo via pipx, and the first line of the script looks like this:
```
#!C:\Users\bmz3\.local\pipx\venvs\git-filter-repo\Scripts\python.exe
```
I did check that the mentioned python is the correct version.

Calling `git-filter-repo` directly works. I guess the python version that is called by git (which is 2.7 here) interferes with the one that should be used. I put a `print(sys.executable)` at the beginning of the script (after modifying line 3239 to make it run), which somehow confirms this: output is `C:\Python27\python.exe`.

qwerbzuio:
@newren Shall I open a new issue for my observations above? Or do you want to reopen this issue? Or keep it closed...?

PR #225, opened 2021-03-16
Improve installation documentation for Windows users

ViliusS:
Since Windows do allow to install Python more easily these days and takes care of most PATH stuff, use recommendations from Microsoft.

The only hiccup is that Git for Windows cannot parse get-filter-repo SHEBANG value when the script is installed via pip. That value is very long by default. Changing it to #!\python.exe solves the issue.

Tested under clean installation of Windows 10 Enterprise version 20H2.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,6 +17,19 @@ no interest in doing so); https://repology.org tracks who packages
 these versions.
 
 
+# Notes for Windows Users
+
+It seems that the default python on Windows is not fully functional,
+and windows might not allow executing python scripts by default.  See
+[issue 124](https://github.com/newren/git-filter-repo/issues/124),
+[issue 36](https://github.com/newren/git-filter-repo/issues/36), and
+[this git mailing list
+thread](https://lore.kernel.org/git/nycvar.QRO.7.76.6.2004251610300.18039@tvgsbejvaqbjf.bet/)
+for details and workarounds.  I believe that Windows users installing
+git-filter-repo via Scoop will be okay, but those using a manual
+installation or pip install may run into these issues.
+
+
 # Manual Installation
 
 filter-repo only consists of a few files that need to be installed:
@@ -70,14 +83,26 @@ filter-repo only consists of a few files that need to be installed:
     instructions regardless of whether the html version of help is
     installed.
 
-So, installation might look something like
-```
-cp -a git-filter-repo $(git --exec-path)
-cp -a git-filter-repo.1 $(git --man-path)/man1
-cp -a git-filter-repo.html $(git --html-path)
-ln -s $(git --exec-path)/git-filter-repo \
-    $(python -c "import site; print(site.getsitepackages()[-1])")/git_filter_repo.py
-```
+So, installation might look something like the following:
+
+1. If you don't have the necessary documentation files (because you
+   are installing from a clone of filter-repo instead of from a
+   tarball) then you can first run:
+
+   `make snag_docs`
+
+   (which just copies the generated documentation files from the
+   `docs` branch)
+
+2. Run the following
+
+   ```
+   cp -a git-filter-repo $(git --exec-path)
+   cp -a git-filter-repo.1 $(git --man-path)/man1
+   cp -a git-filter-repo.html $(git --html-path)
+   ln -s $(git --exec-path)/git-filter-repo \
+       $(python -c "import site; print(site.getsitepackages()[-1])")/git_filter_repo.py
+   ```
 
 # Installation via [pip](https://pip.pypa.io/)
 
@@ -85,6 +110,12 @@ For those who prefer to install python packages via pip, you merely need
 to run:
 
     $ pip3 install git-filter-repo
+
+However, the place where pip places that package might not be in your
+$PATH (thus requiring you to manually update your $PATH afterwards),
+and on windows the pip install might not take care of python-specific
+issues for you (see "Notes for Windows Users", above).  As such,
+installation via package managers is recommended instead.
 
 
 # Installation via Makefile
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,8 +19,33 @@ these versions.
 
 # Notes for Windows Users
 
-It seems that the default python on Windows is not fully functional,
-and windows might not allow executing python scripts by default.  See
+If you don't do a full blown Python web development on Windows the
+best way to install git-filter-repo is to use Python and pip from
+[Microsoft
+Store](https://docs.microsoft.com/en-us/windows/python/beginners).
+Follow these simple steps:
+
+1. Install Python from Microsoft Store.
+2. Run `pip install git-filter-repo` from the Windows Command Prompt.
+3. Copy git-filter-repo file from
+    *%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\LocalCache\local-packages\Python39\Scripts*
+    to the directory pointed to by `git --exec-path`,
+
+    or
+
+    add
+    *%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\LocalCache\local-packages\Python39\Scripts*
+    to you PATH environment variable.
+    
+    The folder above could be different depending on your Python's
+    version.
+4. Change first line of the git-filter-repo script to
+`#!\python.exe`.
+
+If you have Python installed via
+[WSL](https://docs.microsoft.com/en-us/windows/python/web-frameworks)
+or from official [Python for Windows
+installer](https://www.python.org/downloads/windows/) see
 [issue 124](https://github.com/newren/git-filter-repo/issues/124),
 [issue 36](https://github.com/newren/git-filter-repo/issues/36), and
 [this git mailing list
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -39,8 +39,6 @@ Follow these simple steps:
     
     The folder above could be different depending on your Python's
     version.
-4. Change first line of the git-filter-repo script to
-`#!\python.exe`.
 
 If you have Python installed via
 [WSL](https://docs.microsoft.com/en-us/windows/python/web-frameworks)
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1862,7 +1862,7 @@ EXAMPLES
                "part of git history, historical versions of the file will "
                "be ignored; only the current contents are consulted."))
     people.add_argument('--use-mailmap', dest='mailmap',
-        action='store_const', const='.mailmap',
+        action='store_const', const=b'.mailmap',
         help=_("Same as: '--mailmap .mailmap' "))
 
     parents = parser.add_argument_group(title=_("Parent rewriting"))


newren:
Is this obsolete with @dscho's changes to Git for Windows to make it accept the Windows Store version of python?  (See comments in #231.)

ViliusS:
Sadly no. I still cannot run `git filter-repo` without changing shebang value to #!\python.exe or #!/usr/bin/python under latest Git for Windows 2.31.1.

I can run it via `python git-filter-repo` though, so it must be something in Git for Windows indeed.

dscho:
Does it work for you if you execute the command in Git Bash?

dscho:
Also, does `python3 ...` work?

ViliusS:
Nope.

```
username@PC MINGW64 ~
$ git filter-repo
error: cannot spawn git-filter-repo: Exec format error
```

To summarize, these shebang values work:
`#!\python3.9.exe`
`#!\python3.exe`
`#!\python.exe`  (all of these do work with and without .exe extension which is standard under CMD)
`#!/usr/bin/python`
`#!/python` (which is probably the same as \python)
`#!/usr/bin/env python` (which probably uses msys path translation to resolve to python.exe)

And these do not work:
`#!C:\Users\username\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\python.exe` (<-- default when installed with pip from MS Store)
`#!C:\Users\username\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\python3.exe`
`#!/C/Users/username/AppData/Local/Microsoft/WindowsApps/PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0/python.exe`

dscho:
It's not msysgit.

Anyway, what do `where python` and `where python3` say in CMD?

dscho:
And: do `python -h` and `python3 -h` work in your Git Bash?

ViliusS:
```
> where python
C:\Users\username\AppData\Local\Microsoft\WindowsApps\python.exe
```
(which is a reparse point and points to C:\Users\username\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\python.exe same for python3)

python -h, using python interactively and `python git-filter-repo` works so that's not an issue with PATHs or permissions. Shebang value has 96 character limit when running from Git for Windows for some reason.

dscho:
Maybe this is the limit right there: https://github.com/git-for-windows/git/blob/8b8cd53ead5e816b846ab5d26052afa3f4d62c13/compat/mingw.c#L1484

ViliusS:
I don't know enough about C to understand why the real limit is 96 and not 100, but yea, it could be it.

dscho:
What I really don't understand is that the hash-bang line should read `#!/usr/bin/env python3`, i.e. it should resolve the interpreter to `env.exe` (which is in `C:\Program Files\Git\usr\bin\env.exe`, a path well below that limit).

BTW I broke down and interrupted my AFK to play with this a bit, and I can finally reproduce your `Exec format error`. I think it will be better to fix this on Git for Windows' side than on `git-filter-repo`'s. Hopefully I'll find some time to play more with that, for now I have to go back to AFK.

In the meantime, could I ask you to figure out an _easy_ way to add the script location to the `PATH` in `~/.bashrc`? That would make much more sense than to copy `git-filter-repo` into Git's exec-path, as the latter method is prone to let the copied version go stale upon `git-filter-repo` upgrades.

ViliusS:
Shebang value is autogenerated by the packager which is used to install git-filter-repo which in my case is pip from the MS Store. I'm afraid we won't have control over it.

I do agree that this needs to be fixed in Git for Windows, however git-filter-repo documentation still needs to be improved and fixing shebang in Git for Windows only removes step 4 in my instructions, so this pull request still stands.

Regarding adding Scripts folder to PATH, I thought about it for some time but could not find an easy way make it fully automated. If we copy git-filter-repo to Git's exec-path it won't be upgraded by the next `pip upgrade git-filter-repo` and if we add that folder to Windows PATH it does contain hardcoded Python version which could break on next Python upgrade, e.g. 3.9 -> 3.10.

Adding script folder to .bashrc still leaves us with broken path with next Python upgrade and only works when running Git for Windows with Bash, not cmd. I believe the correct way would be to contact Python Software Foundation to see if they can implement this during MS Store installation, however I didn't find an easy way to leave this feedback for them.

dscho:
Maybe `printf '#!/bin/sh\n\nexec python git_filter_repo "$@"\n' >~/bin/git-filter-repo` would still be the best solution.

ViliusS:
I doubt that these instructions are easier than to copy to Git's exec path :) But maybe it's just me.

newren:
So...

I'm not particularly happy with the really lengthy instructions.  They suggest there is a more fundamental issue that needs to be solved elsewhere, and since I have multiple reports of people using filter-repo successfully on Windows, I'm worried the instructions might break some of those users while fixing it for others.  If it were instead another issue or something I could link to that people could check, or if we could root cause that it really affected everyone, then I'd be more at ease that these instructions wouldn't break others.

But fundamentally, there are a few more issues with these instructions:

1. The suggestion that installation should be done via pip (as opposed to could be done) is out of place.  The very first people to package git-filter-repo for Windows and submit things to fix up git-filter-repo were the Scoop folks.  They were faster than nearly all the linux packaging folks, in fact.  And I've never heard a complaint from anyone installing via the Scoop package manager on Windows that things didn't just work.  Yet I've heard multiple complaints from folks using pip, making me nearly regret even suggesting pip as a possibility.  I moved it from the top of the page (as put there by the folks who did the pip packaging to make it possible) to near the bottom already, maybe I need to bury it even more.

2. I don't think explaining how to install dependencies belongs in the scope of filter-repo.  (The Python project should provide instructions on how to install python; not me.  Also, if I'm adding instructions on how to install python, on any platform, I probably need to start adding it for other platforms.  Those will probably need lots of updates over time and I'm putting myself into a losing game). So steps 1 & 2 do not belong here.

3. The instructions suggest pip might be buggy.  If so, why is it even being suggested?  In particular, it suggests both that pip isn't sticking the script in $PATH (in which case is that Windows-specific or all platforms?  If all platforms, it doesn't belong in the windows specific section.  It looks like I already noted that problem under pip, so repeating it in the windows specific section seems duplicative, so step 3 really doesn't fit either).  The instructions also suggest that the shebang mangling pip is doing might be incorrect (though Johannes suggests the second issue might be a Git for Windows bug rather than a pip one).

I think all that's left after the above is (a) a fix for the shebang handling in Git for Windows, and (b) maybe being more overt on the HERE-BE-DRAGONS warning in the pip section about how it can cause problems for old versions of Git for Windows (anything released prior to April 2021), and additionally that it leaves to the users one of the installation steps -- copying the script into $PATH -- and that perhaps folks should strongly consider using other package managers instead.

dscho:
> Maybe this is the limit right there: https://github.com/git-for-windows/git/blob/8b8cd53ead5e816b846ab5d26052afa3f4d62c13/compat/mingw.c#L1484

I got a chance to try out this hypothesis. Increasing the size from 100 to 248 fixes the `Exec format error` bug.

dscho:
> Increasing the size from 100 to 248 fixes the `Exec format error` bug.

I opened https://github.com/git-for-windows/git/pull/3165 to address that.

newren:
> > Increasing the size from 100 to 248 fixes the `Exec format error` bug.
> 
> I opened [git-for-windows/git#3165](https://github.com/git-for-windows/git/pull/3165) to address that.

Awesome, thanks!

ViliusS:
> So...
> 
> I'm not particularly happy with the really lengthy instructions. They suggest there is a more fundamental issue that needs to be solved elsewhere, and since I have multiple reports of people using filter-repo successfully on Windows, I'm worried the instructions might break some of those users while fixing it for others. If it were instead another issue or something I could link to that people could check, or if we could root cause that it really affected everyone, then I'd be more at ease that these instructions wouldn't break others.

I will answer your other points in a minute, but let me first explain my experience as a git-filter-repo user.

Current note for Windows users in install docs suggests that there is something wrong with "default python" on Windows. OK, so I went to Python.org and tried to install it from there. I have also tried to download git-filter-repo from this repository and just place the file in my PATH as suggested. This didn't work. I have searched several rabbit holes coming empty handed. Since there are multiple ways how Python can be installed on Windows from the linked issues I had to guess if that relates to my setup or not (most of the time it didn't). After digging for several hours, to the point where I forgot why do I need Python anyway, I went back to using `git filter-branch`.

I would never came back if not for gitlab.com which support rewriting history only with git-filter-repo. This time I thought that I won't follow git-filter-repo documentation but just install everything as I always do, using standard Windows tools. First I had to find out that, even if documentation suggest so, there is **no** default python on Windows. OK, so I thought, let's try to install it. What is the best way? Naturally I consulted my OS documentation which pointed to MS Store or WSL installs. MS Store install was a breeze and I have confirmed that python was placed in my PATH via `python --version` etc. Documentation specifically states that Python from MS Store comes with pip which is the default package manager for Python, so I have used it to install git-filter-repo. Now running `git filter-repo` still could not find it but I have remembered this line `git-filter-repo is a single-file python script, which was done to make installation for basic use trivial: just copy it into your $PATH.` So I dig where pip put the file and added it to my PATH. Everything worked and it only took me 10 minutes instead of several our when following documentation and reading through various issues.

Naturally, I was frustrated because of this and thought that nobody should go through the same experience as me, hence this PR.
 
> But fundamentally, there are a few more issues with these instructions:
> 
> 1. The suggestion that installation should be done via pip (as opposed to could be done) is out of place.  The very first people to package git-filter-repo for Windows and submit things to fix up git-filter-repo were the Scoop folks.  They were faster than nearly all the linux packaging folks, in fact.  And I've never heard a complaint from anyone installing via the Scoop package manager on Windows that things didn't just work.  Yet I've heard multiple complaints from folks using pip, making me nearly regret even suggesting pip as a possibility.  I moved it from the top of the page (as put there by the folks who did the pip packaging to make it possible) to near the bottom already, maybe I need to bury it even more.

I can change it from "should" to "could" and your point about Scoop folks is probably correct, but I'm thinking this way: software under **every** OS should be installed via standard way as much as possible, be it rpm/deb on Linux, dmg on macOS, or MS Store/MSI on Windows. This way it automatically takes cares of OS specific issues, most of the time is auto-updated, provisioned, etc. etc. Not every developer uses Python for their work, I would argue that a fair amount of them doesn't even know what Python or pip is. Also, if you are HTML developer, or PHP developer, or C# developer on Visual Studio, or database developer, you probably won't have **any** of the package managers installed (I didn't). Requiring these folks to install separate package manager just to install one simple python script to be able to rewrite git history is insane. 

If you are afraid that these instructions will break something, you can always revert them (using git-filter-repo to remove them from existence :)), but facts show that folks are still stuggling without them, jumping from one install method to another without any good reason, https://github.com/newren/git-filter-repo/issues/193 https://github.com/newren/git-filter-repo/issues/189 https://github.com/newren/git-filter-repo/issues/124 Some of the issues they discovered are already fixed in latest Windows/Git for Windows versions, others are just not understanding how various Python install alternatives work and which PATHs are already defined and should not be touched. By the way most of these links suggest that installing from MS Store finally solved their problems.

> 2. I don't think explaining how to install dependencies belongs in the scope of filter-repo.  (The Python project should provide instructions on how to install python; not me.  Also, if I'm adding instructions on how to install python, on any platform, I probably need to start adding it for other platforms.  Those will probably need lots of updates over time and I'm putting myself into a losing game). So steps 1 & 2 do not belong here.

I still think one or two lines how to install dependencies won't hurt and will solve more issues than it produces. They are pointing to official documentation and if something changes MS will take care of it. Even if MS Store method won't be recommended some time in the future others (or me) could make another PR to update documentation, that's natural cycle of open source software. The same goes for other OSes, if anybody is willing to provide a PR to make more detailed instructions, why not?

> 3. The instructions suggest pip might be buggy.  If so, why is it even being suggested?  In particular, it suggests both that pip isn't sticking the script in $PATH (in which case is that Windows-specific or all platforms?  If all platforms, it doesn't belong in the windows specific section.  It looks like I already noted that problem under pip, so repeating it in the windows specific section seems duplicative, so step 3 really doesn't fit either).  The instructions also suggest that the shebang mangling pip is doing might be incorrect (though Johannes suggests the shebang handling might be a Git for Windows bug rather than a pip one).

That's not correct. pip is not buggy, Python itself doesn't register Scripts in the PATH, but I'm sure that will be solved sooner rather then later. I just need to find a way to reach Python folks to leave this feedback for them.
 
> I think all that's left after the above is (a) a fix for the shebang handling in Git for Windows, and (b) maybe being more overt on the HERE-BE-DRAGONS warning in the pip section about how it can cause problems for old versions of Git for Windows (anything released prior to April 2021), and additionally that it leaves to the users one of the installation steps -- copying the script into $PATH -- and that perhaps folks should strongly consider using other package managers instead.

I will test it when the next Git for Windows version is released, share my experience and update the instructions if that's OK?

dscho:
I think that it is a very valid point that it's too hard to install `git filter-repo` on Windows.

Maybe we can do something about it? E.g. ship a `git-filter-repo` _shell_ script that detects whether Python is installed (or actually, whether `pip.exe` is in the `PATH`, and if not, exits with error after opening Windows Store on the Python page. If it _is_ installed, the script could call `pip.exe show -f git-filter-repo` to find out the path of the actual script and `exec` it, or install the package if the command fails because it's not installed yet.)

ViliusS:
I don't think that it's too hard from the perspective of git-filter-repo. PIP or any other package manager should be fine, because folks using git should be accustomed with CLI tools anyway.

Even if git-filter-repo could provide a shell script, the main problem remains, i.e. finding a good way to install Python itself. I think we just need to guide folks a little bit. MS Store install is definitely a step in the right direction, IMHO.

If your fix solves shebang issue and if there is a way for MS Store install to add Scripts folder to the PATH it would be perfect. I have dropped an email to @zooba to see if anything could be done regarding that.

dscho:
I meant shipping said script in Git for Windows.

ViliusS:
OK, so I did couple of experiments on clean Windows account.

1. Download https://github.com/newren/git-filter-repo/blob/main/git-filter-repo manually and place it in my PATH or Git's exec-path.
2. Then:
    ```
    C:\Users\test>git filter-repo
    Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > 
    Manage App Execution Aliases.
    ```
3. Install Python from MS Store.
4. Then:
    ```
    C:\Users\test>git filter-repo
    No arguments specified.
    ```

It works, no admin permissions needed! The experience is more or less OK actually.

> I meant shipping said script in Git for Windows.

I'm not sure if there is a clean way to "add" git-filter-repo for Git's execution, but maybe your idea is not so bad. To give more control to newren regarding installation method, the PowerShell script could initiate Invoke-WebRequest('https://github.com/newren/git-filter-install.py') installer which would automatically show the error message above, if Python is not installed. If it is installed it could try to install git-filter-repo using pip and return the path. The same install script could even detect OS and install everything accordingly.

dscho:
> [...] and placed it in my PATH or Git's exec-path.

I am still a bit unhappy at how laborious this is (you have to look at `pip show -f git-filter-repo` to see the location, and then _still_ munge it a bit because of the `..\Scripts` before you can adjust the `PATH`, and most Windows users still have no idea how they can set their `PATH` so that it is persisted across reboots).

ViliusS:
> > [...] and placed it in my PATH or Git's exec-path.
> 
> I am still a bit unhappy at how laborious this is (you have to look at `pip show -f git-filter-repo` to see the location, and then _still_ munge it a bit because of the `..\Scripts` before you can adjust the `PATH`, and most Windows users still have no idea how they can set their `PATH` so that it is persisted across reboots).

Probably true, however if you download get-filter-repo manually you can place it *anywhere* in a PATH without modifying it. Let's say in C:\Program Files\Git\cmd It won't be automatically updated, but again, this is for manual install.

If you want installation to be completely automated, then I'm afraid that is not possible without your mentioned shell script. Or thinking about this further, why git-filter-branch is included in default git install, but git-filter-repo is not? :)

dscho:
Heh, there is talk on the Git mailing list to kick `git filter-branch` out from the tree...

ViliusS:
I can confirm that step 4 from my instructions is not needed now as it is fixed in Git for Windows 2.32.0.

I have pushed new commit regarding this but noticed that INSTALL.md was updated few days ago. Should we still keep this PR and the discussion open?

newren:
Great!  I'm glad step 4 is no longer necessary; that's useful information.  Is step 3, the step about copying the script, still needed?  If so, does #217 fix it (should I make a new release with that fix)?

Your steps 1 & 2 are still too prescriptive, ignoring Scoop and thus suggesting pip is *the* way to install git-filter-repo on Windows.  I'm not willing to accept that, especially since the Scoop folks provided reports and packaging assistance very early on (incidentally, they are almost always the fastest to package new releases too).

I did update INSTALL in order to mention dscho's work to fix various Windows things, and to try to address several of the different issue reports.  I would have preferred to be able to mention for sure whether editing the shebang was still needed, to be certain whether I had listed enough steps (e.g. is copying still needed), but I don't have access to a Windows machine to test and thus just don't know the answers.  So I had to use the ugly "well, maybe this will help" style of wording, and then link to lots of tickets for more discussion -- including listing this one first since you did some really good investigative work.  Do the current instructions look sufficient to you?  If so, yeah we can close.  If not, well let's keep this open and try to address those.

Thanks for all your help so far!

ViliusS:
> Great! I'm glad step 4 is no longer necessary; that's useful information. Is step 3, the step about copying the script, still needed? If so, does #217 fix it (should I make a new release with that fix)?

It looks like different issue. As I understood from the information given in #144 he had two Pythons installed. From the looks of it, because of C:\Python2.7 path, one Python was installed from python.org, another by other means. Since pipx creates its own venvironment the patch just chooses correct Python version which is probably not related to my installation method. Step 3 is needed not because git cannot find the correct Python, but because it cannot find git-filter-repo itself.

Anyway, I'm unsure how to test this without version release on pip. If we want to test it on installation method in the instructions I have to install git-filter-repo from pip.

I will take a look at the updated documentation later. As far as I glanced it is definitely an improvement, but there are some errors, like the information that Windows 10 already comes with Python installed. It doesn't.

ViliusS:
Oh and I forgot to mentioned that I have reached to Microsoft regarding PATH issue. My question and Steve's, who works on Python at MS, response below. Not really helpful, but I thought I leave this information here.

> > I‘m using Python on Windows from MS Store and have to say that your work is great! Love the fast install.
> > 
> > I have one issue though. Apparently Scripts folder (%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\LocalCache\local-packages\Python39\Scripts) is not automatically added to the PATH. This produces issues when trying to run python scrips which are installed via pip. For example this project https://github.com/newren/git-filter-repo is a python script which needs to be run by Git for Windows. In order for “git filter-repo” to work as a git command this script needs to be in the PATH or in git’s execution path.
> > 
> > I know how to add it to the PATH manually, but do you think this could be taken care off during Python install? That would be great! Or maybe there is a way to handle this gracefully from the git-filter-repo’s perspective?
> 
> I’m afraid there’s no mechanism in Store package installs to modify environment variables (by design – the install format is intended to prevent doing things that break other apps or cannot be easily reverted), and there really isn’t a better option than “python -m …” for launching scripts right now. Perhaps git-filter-repo can use that approach?
> 
> The model of a Python installation being the “container” for a bunch of global scripts isn’t really a great design anyway (particularly for Windows), and venv’s still work fine, so this hasn’t been showing up as a critical issue. But still, I’m working on getting some OS-level limitations changed that will make it easier to do. I’d suggest that a self-contained app should contain its own copy of Python (most likely the embeddable distro), rather than expecting a user to install two separate things. But that can be a lot of complexity for you the developer, so I understand why many don’t bother.
> 
> Sorry that isn’t a more helpful response. Hopefully you can figure out a way to make it work.

newren:
> I will take a look at the updated documentation later. As far as I glanced it is definitely an improvement, but there are some errors, like the information that Windows 10 already comes with Python installed. It doesn't.

That page doesn't claim that Windows 10 comes with Python installed; you've unfortunately made three errors in translation of what the page says (inserting a version of Windows, changing Python-like to Python, and dropping the "It has been reported that" context).  It looks like my error was in translating "stub" to "stripped down version", maybe I should correct that; see https://github.com/newren/git-filter-repo/issues/36#issuecomment-568933825.  Maybe I should just link to that comment so people can see the report.

newren:
> Oh and I forgot to mentioned that I have reached to Microsoft regarding PATH issue. My question and Steve's, who works on Python at MS, response below. Not really helpful, but I thought I leave this information here.

Thanks for checking.  But it looks like it's not really a Microsoft issue; it's apparently a problem inherent to using pip.  See https://stackoverflow.com/questions/35898734/pip-installs-packages-successfully-but-executables-not-found-from-command-line, and note that most the comments are about Mac OS X, though there are several for Linux, and even one for Windows, with folks providing workarounds/solutions/answers for all three platforms.


In my INSTALL.md, I previously had this text under the pip section:

> However, the place where pip places that package might not be in your $PATH (thus requiring you to manually update your $PATH afterwards)

but I think that wasn't scary or big enough.  So I just replaced it with some text at the top of the pip section that says:

> This method is NOT recommended, because installing via pip does not place the package in your $PATH. If you go this route, you have to copy the file into an appropriate directory (either `git --exec-path` or something else in $PATH), or modify your $PATH to include the directory where pip places things, but you could have just done either of those things after cloning filter-repo directly so this installation method provides virtually no benefit.

and linked to the above stackoverflow page.  I made this change in commit fd64b0c7c0c2 (INSTALL.md: more clarifications and links, 2021-06-11)

CS-Cowboy:
@OP You ought to add a new entry to your PATH and point it at the git-filter-repo zip file (or clone) you got from Github. From then on you will be able to call git filter-repo from the command line on Windows 10 (or powershell). 

While I agree the documentation is lengthy and imprecise, I think you're really far off track.

If you have any more than one python installation that will cause problems. MSYS64, CMAKE, some CPP tools for windows do create their own Python installations and this **will** cause problems for you. 
Also you can obtain Python from the Windows 10 store, which can also cause conflicts.

So I think its a good idea to remove extraneous Python installations and try adding that entry to your Path. 
(Below: adding the PATH environment variable entry)
Step_2
(Below: testing git filter-repo --analyze )
Step_3

newren:
I've made a few changes since this PR.  I think the most important one is just telling people to download the `git-filter-repo` script, then run `python3 git-filter-repo ...`.  Cause that's all you really have to do, and trying to get `git filter-repo` magic to work (where git looks for git-filter-repo in PATH, then invokes it, then the OS needs to recognize the shebang and hope it works) is just too advanced for poor Windows to handle apparently.  See https://github.com/newren/git-filter-repo/blob/doc-update/INSTALL.md#simple-installation  However, there's a big [Notes for Windows users](https://github.com/newren/git-filter-repo/blob/doc-update/INSTALL.md#notes-for-windows-users) section too.

PR #239, opened 2021-04-17
git-filter-repo.txt: fix typo in paths-from-file example

Codym48:
The "Filtering based on many paths" section includes this code snippet,
```
# An example that is like --path-regex:
regex:^.*/.*/[0-9]{4}-[0-9]{2}-[0-9]{2}.txt$
```
and this  text, 

> files whose name was of the form YYYY.MM-DD.txt at least two subdirectories deep

Update the text to YYYY-MM-DD.txt to match the regex in the code snippet.

### DIFF
--- a/Documentation/git-filter-repo.txt
+++ b/Documentation/git-filter-repo.txt
@@ -761,7 +761,7 @@ git filter-repo --paths-from-file stuff-i-want.txt
 
 to get a repo containing only the toplevel README.md file, the guides/
 and tools/releases/ directories, all python files, files whose name
-was of the form YYYY.MM-DD.txt at least two subdirectories deep, and
+was of the form YYYY-MM-DD.txt at least two subdirectories deep, and
 would rename tools/ to scripts/ and rename files like foo/bar/baz.text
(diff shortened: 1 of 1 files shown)


newren:
Sweet, thanks for catching this!

Could you add your Signed-off-by to the commit message, and use a real email address for recording the commit?

Also, if you could prepend the first line of the commit message with `git-filter-repo.txt: `, that'd be great, but I'd be happy to fix that up for you if you can add the Signed-off-by and fix the email address.

newren (changes requested):
Marking as needing changes for the Signed-off-by and email addresses.

Codym48:
Today I learned about the Signed-off-by trailer (https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches#n305) and the SCO lawsuits that preceded it (https://stackoverflow.com/a/1962112/2597078) and read the rest of the contribution guidelines and tried to do some line wrapping in the process.

Thanks for your patience

newren (approved):

newren:
Thanks for the contribution!

PR #241, opened 2021-04-17
Fix "fatal: this operation must be run in a work tree" error when running bfg-ish

### LINKED ISSUE #240: "fatal: this operation must be run in a work tree" error message when running bfg-ish
This error message is printed whenever I run `bfg-ish` (no matter what the working directory is, including in the GIT_WORK_TREE).

```command-output
$ LANG=C PYTHONPATH="${PWD}" ./contrib/filter-repo-demos/bfg-ish --replace-text replace.txt test/.git
Processed 3 blob sizes
Parsed 1 commits
New history written in 0.01 seconds...
fatal: this operation must be run in a work tree
Completely finished after 0.02 seconds.
Repacking your repo and cleaning out old unneeded objects
Enumerating objects: 3, done.
Counting objects: 100% (3/3), done.
Writing objects: 100% (3/3), done.
Total 3 (delta 0), reused 3 (delta 0), pack-reused 0
```

The error message doesn't seem to be critical though, but I still like to eliminate it whenever possible.

Lin-Buo-Ren:
Hardcoding GIT_WORK_TREE to GIT_DIR seems to be a sane way to silence the error.  Please review.

Fixes #240.

Signed-off-by: 林博仁(Buo-ren Lin) <Buo.Ren.Lin@gmail.com>

### DIFF
--- a/contrib/filter-repo-demos/bfg-ish
+++ b/contrib/filter-repo-demos/bfg-ish
@@ -371,6 +371,7 @@ class BFG_ish:
     bfg_args = self.parse_options()
     preserve_refs = self.get_preservation_info(bfg_args.preserve_ref_tips)
 
+    os.environ["GIT_WORK_TREE"] = os.path.abspath(os.fsdecode(bfg_args.repo))
     os.chdir(bfg_args.repo)
     bfg_args.delete_files = java_to_fnmatch_glob(bfg_args.delete_files)
     bfg_args.delete_folders = java_to_fnmatch_glob(bfg_args.delete_folders)


Lin-Buo-Ren:
Holding up as the patch doesn't handle relative path given case, and, does not handle Windows-compatibility.

Lin-Buo-Ren:
> Holding up as the patch doesn't handle relative path given case, and, does not handle Windows-compatibility.

Fixed.

newren:
Thanks for sending in patches, and sorry again for taking a while to respond.

In this particular case, I'd rather just tell the user that they are specifying the repo erroneously.  When the repository is not bare, they should specify the toplevel directory.  I just pushed a change to print such a message; see commit 2f306c6d3f09 (bfg-ish: add some sanity checks on the specified repo, 2021-06-05)

PR #242, opened 2021-04-17
Fix relative path compatibility for --replace-text and bfg_args.repo

### LINKED ISSUE #203: bfg-ish fails for repo arguments other than '.'
1) bfg-ish apparently assumes at some places to have been invoked from inside repo.

2) Also, where bfg-ish cds into the repo dir, the argument files (e.g. replacement.txt) are searched with relative paths from inside the repo, not relative to bfg-ish invocation cwd, which is unexpected. But even when compensating this by passing ../replacements.txt the command fails because of 1)

```
bfg-ish -fi app -fi '*.{rb}' --replace-text ../replacements.txt myrepo
Processed 1572 blob sizes
Parsed 104 commits
New history written in 0.92 seconds...
HEAD is now at af86408 whitelist 10.250.0.1 for web_console
Completely finished after 0.93 seconds.
Repacking your repo and cleaning out old unneeded objects
Traceback (most recent call last):
  File "/home/user/src/git-filter-repo/contrib/filter-repo-demos/bfg-ish", line 439, in <module>
    bfg.run()
  File "/home/user/src/git-filter-repo/contrib/filter-repo-demos/bfg-ish", line 435, in run
    fr.RepoFilter.cleanup(bfg_args.repo, repack=True, reset=need_another_reset)
  File "/home/user/src/git-filter-repo/contrib/filter-repo-demos/git_filter_repo.py", line 2973, in cleanup
    subproc.call(cmd, cwd=repo)
  File "/usr/lib/python3.7/subprocess.py", line 323, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib/python3.7/subprocess.py", line 775, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.7/subprocess.py", line 1522, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: b'myrepo': b'myrepo'

```

Tested with version e9e0308df1e9e792598b8e41d0086d7a785e9cc5

Lin-Buo-Ren:
Making the paths absolute when it is not to avoid FileNotFound errors due to the working directory change.

Fixes #203.

Signed-off-by: 林博仁(Buo-ren Lin) <Buo.Ren.Lin@gmail.com>

### DIFF
--- a/contrib/filter-repo-demos/bfg-ish
+++ b/contrib/filter-repo-demos/bfg-ish
@@ -371,6 +371,7 @@ class BFG_ish:
     bfg_args = self.parse_options()
     preserve_refs = self.get_preservation_info(bfg_args.preserve_ref_tips)
 
+    work_dir = os.getcwd()
     os.chdir(bfg_args.repo)
     bfg_args.delete_files = java_to_fnmatch_glob(bfg_args.delete_files)
     bfg_args.delete_folders = java_to_fnmatch_glob(bfg_args.delete_folders)
@@ -395,6 +396,9 @@ class BFG_ish:
       extra_args += ['--preserve-commit-hashes']
     new_replace_file = None
     if bfg_args.replace_text:
+      if not os.path.isabs(bfg_args.replace_text):
+        bfg_args.replace_text = os.path.join(work_dir, bfg_args.replace_text)
+
       new_replace_file = self.convert_replace_text(bfg_args.replace_text)
       rules = fr.FilteringOptions.get_replace_text(new_replace_file)
       self.replacement_rules = rules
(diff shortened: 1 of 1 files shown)


Lin-Buo-Ren:
Holding up as the patch doesn't handle Windows-compatibility.

Lin-Buo-Ren:
> Holding up as the patch doesn't handle Windows-compatibility.

Done.

newren:
Thanks for sending in patches!  Sorry for taking a while to respond.

This patch looks great.  I would much rather the commit message explained the problem being fixed than including text like "Fixes #203" which causes people looking at the code to have to dig up external references...especially since those references may not be permanent (there's no guarantee I'll always use github to host git-filter-repo).  However, you described the problem well, so I just fixed up the commit message and pushed it up as commit e732141363fc (Fix relative path compatibility for --replace-text and bfg_args.repo, 2021-04-17)

PR #252, opened 2021-05-07
filter-repo: Use author date instead of commit date in --analyze

CervEdin:
The date displayed in the analysis is the *commit date*. This date is often less accurate when analyzing "when" something happened, since it will be reset by things like cherry-picks, rebase etc. 

In my experience, the *author date* provides a more reliable indication.

However, I must concede that I'm not an expert on the subject. It could very well be that the *commit date* is a lot more reasonable. In my particular case however, the *commit date* is effectively useless as it presents several years of changes having occurred within the last two days.

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -2319,7 +2319,7 @@ class RepoAnalyze(object):
     commit_parse_progress = ProgressWriter()
     num_commits = 0
     cmd = ('git rev-list --topo-order --reverse {}'.format(' '.join(args.refs)) +
-           ' | git diff-tree --stdin --always --root --format=%H%n%P%n%cd' +
+           ' | git diff-tree --stdin --always --root --format=%H%n%P%n%ad' +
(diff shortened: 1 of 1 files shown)


newren (changes requested):
Yeah, I actually mulled it over for a while.  There are definitely usecases where author date would be better, and others where committer date is better.  Since `git log` and log-like commands default to using committer date (e.g. with --since, --after, --until, --before) I think it's the right default.

It'd probably be reasonable to add an option to work with author date, though.

If you'd like to do that, please also amend the commit message to include your Signed-off-by trailer, as per https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches#n305.  It'd also be nice to re-wrap the commit message at 72 characters, though I'm happy to fix that up if you have your Signed-off-by in there.

newren:
No response, so I'll go ahead and close for now.  Feel free to reopen if you'd like to work on this.

PR #275, opened 2021-08-22
Add support to set directory for --analyze report

rndbit:
Please consider this minor feature to allow setting a custom directory for writing result of --analyze.

While preparing a huge repo for migration I found useful to keep a pristine cloned copy read only, but still wanted to be able to run --analyze report on it, except that git-filter-repo --analyze wants to write inside git repo. This trivial change helps in this case. I can now invoke it as

    git --git-dir=/huge-RO.git filter-repo --analyze --report-dir ~/huge-repo-analisis

Thanks for reading, and for the great tool.


rndbit:
Closing this PR to replace with equivalent by with commits including "Signed-off-by"

PR #277, opened 2021-08-24
Calculate the version from the module, not the entry_point

stefanor:
When git-filter-repo is installed, sys.argv[0] will be an entry-point stub, not the relevant Python module.


newren:
Thanks!

PR #279, opened 2021-08-26
Fix issue 276 binary blob detection

rndbit:
Hi,

Please consider a fix and tests for issue 276 re binary blob detection not working and --text-replace replaces text in binary files which is not intended.

Thanks.


rndbit:
Closing this PR, to fix commit messages to add "Signed-off-by", will create a new PR with changes

PR #280, opened 2021-09-01
Bugfix/issue 276 fix binary blob detection

rndbit:
Hi,

Please consider a fix and tests for issue 276 re binary blob detection not working and --text-replace replaces text in binary files which seems was not intended.

Thanks.

### DIFF
--- a/t/t9390-filter-repo.sh
+++ b/t/t9390-filter-repo.sh
@@ -850,6 +850,87 @@ test_expect_success '--replace-text all options' '
 	)
 '
 
+test_expect_success '--replace-text binary zero_byte-0_char' '
+	(
+		set -e
+		set -u
+		REPO=replace-text-detect-binary
+		FILE=mangle.bin
+		OLD_STR=replace-from
+		NEW_STR=replace-with
+		# used with printf, contains a zero byte and a "0" character, binary
+		OLD_CONTENT_FORMAT="${OLD_STR}\\0${OLD_STR}\\n0\\n"
+		# expect content unchanged due to binary
+		NEW_CONTENT_FORMAT="${OLD_CONTENT_FORMAT}"
+
+		rm -rf "${REPO}"
+		git init "${REPO}"
+		cd "${REPO}"
+		echo "${OLD_STR}==>${NEW_STR}" >../replace-rules
+		printf "${NEW_CONTENT_FORMAT}" > ../expect
+		printf "${OLD_CONTENT_FORMAT}" > "${FILE}"
+		git add "${FILE}"
+		git commit -m 'test'
+		git filter-repo --force --replace-text ../replace-rules
+
+		test_cmp ../expect "${FILE}"
+	)
+'
+
+test_expect_success '--replace-text binary zero_byte-no_0_char' '
+	(
+		set -e
+		set -u
+		REPO=replace-text-detect-binary
+		FILE=mangle.bin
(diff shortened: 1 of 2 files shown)


t/t9390-filter-repo.sh
```diff
@@ -850,6 +850,87 @@ test_expect_success '--replace-text all options' '
...
+test_expect_success '--replace-text text-file no_zero_byte-zero_char' '
+	(
+		set -e
+		set -u
+		REPO=replace-text-detect-binary
+		FILE=mangle.bin
+		OLD_STR=replace-from
+		NEW_STR=replace-with
+		# used with printf, contains no zero byte but contains a "0" character, text
+		OLD_CONTENT_FORMAT="${OLD_STR}\\n${OLD_STR}\\n"
```
newren, line 916: The comment two lines above claims this contains a "0" character, but I  don't see one.  Is the comment wrong, or the line following it?

newren (changes requested):
Sweet, thanks.  And sorry for the delay.

The code changes look good (both the last patch and this one), other than a small issue with a comment noted below.

For the commit messages:
  * Please use a real email address
  * Please don't reference GitHub Issues by number; I want the history to be self-standing if I move the repository elsewhere.
  * Git style is to reference an "area: short description" for the one-line summary, and use full sentences for subsequent paragraphs.
  * It might be worth describing these commits just a bit more.

So, commit message for the first commit would look like:

"""
filter-repo: fix detection of binary blobs for --replace-text

The construct 'for x in b"..."' will iterate over a list of integers,
not a list of 1-character bytestrings, and thus break our 'x == b"\0"'
construct.  Further, we were supposed to be searching for NUL characters
not 0 characters, thus b"\0" instead of b"0".

Signed-off-by: $NAME <$real_email>
"""

and for the second:

"""
Add tests for detecting binary blobs

We previously failed to detect binary files correctly; add some tests to
make sure that regression does not return.

Signed-off-by: $NAME <$real_email>
"""

rndbit:
Hello and thanks for looking into it.

> * Please use a real email address

I'm very opposed to disclosing my real identity with content accessible to entire world and I've got into a bit of effort to obscure my details. Lets call it paranoia. I'll think about a compromise.

> The comment two lines above claims this contains a "0" character, but I don't see one. Is the comment wrong, or the line following it?

Yes, I missed "0" char in that test input, very silly, thanks for spotting it, will fix.

Minor thing, regarding commit message examples you offered, I wish to clarify regarding the use of "I"/"We" in the messages, AFAIK this is not recommended, something to do with "imperative" something. I'm not passionate about this one.

I realised I did not have to close PR to replace branch, just rebase apparently would have been enough. I'm still new-ish to contributing on github.

rndbit:
Hello again. I believe I've addressed the requested changes. Fixed test. Improved comments. Used real email in commit and in sign off.

I've made another branch name `bugfix/text-replacement-skip-binary-blobs` without issue number in the branch name at same commit as branch in this PR. Not sure I want to discard this PR just to change branch name. Let me know if you do.

newren:
Thanks for the contribution!  And my deep apologies for disappearing into a black hole for 2 months.  This is a great fix you provided; filter-repo is better for it.

rndbit:
No need to apologise. I'm carving a huge disorganised repo into projects, it would be impossible without git-filter-repo. Thank you.

PR #281, opened 2021-09-01
Add support to set directory for --analyze report

rndbit:
Please consider this minor feature to allow setting a custom directory for writing result of --analyze.

While preparing a huge repo for migration I found useful to keep a pristine cloned copy read only, but still wanted to be able to run --analyze report on it, except that git-filter-repo --analyze wants to write inside git repo. This trivial change helps in this case. I can now invoke it as

git --git-dir=/huge-RO.git filter-repo --analyze --report-dir ~/huge-repo-analisis
Thanks for reading, and for the great tool.

Signed-off-by: rndbit <43857958+rndbit@users.noreply.github.com>

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1758,6 +1758,12 @@ EXAMPLES
         help=_("Analyze repository history and create a report that may be "
                "useful in determining what to filter in a subsequent run. "
                "Will not modify your repo."))
+    analyze.add_argument('--report-dir',
+        metavar='DIR_OR_FILE',
+        type=os.fsencode,
+        dest='report_dir',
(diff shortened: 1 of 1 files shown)


rndbit:
Updated comment and sign off per requirement discussed in different PR

newren:
This broke one of the regression tests, and added some new lines that were not covered by the line coverage.  But everything else looked good, so I tweaked the commit a bit to include some test changes and pushed as commit e9d5ab3529ed (filter-repo: add option --report-dir to set custom analysis dir, 2021-08-22).

Thanks for the contribution!

PR #284, opened 2021-09-13
lint-history: allow forwarding git-filter-repo options ( e.g. --refs master..feature )

kxrob:
Motivation see: https://github.com/newren/git-filter-repo/discussions/283 : How to run a code formatter on a feature branch only?


fatash89 (approved):

contrib/filter-repo-demos/git-lint-history
```diff
@@ -0,0 +1 @@
+git_lint_history.py
```
newren, line 1: Why rename it, though?

contrib/filter-repo-demos/git_lint_history.py
```diff
@@ -97,9 +98,6 @@ def is_relevant(filename):
               "the file's extension), then pass this argument"))
 parser.add_argument('command', nargs=argparse.REMAINDER,
         help=("Lint command to run, other than the filename at the end"))
-lint_args = parser.parse_args()
-if not lint_args.command:
-  raise SystemExit("Error: Need to specify a lint command")
```
newren, line 102: This doesn't match the commit message as far as I can tell; you're not only making a main() but adding other changes as well.  What's the purpose of those changes?  Can they be split into another commit, so they can be independently motivated in their own commit message?

contrib/filter-repo-demos/git_lint_history.py
```diff
@@ -149,24 +147,37 @@ def lint_non_binary_blobs(blob, metadata):
...
-  cat_file_process.stdin.close()
-  cat_file_process.wait()
-else:
-  filter = fr.RepoFilter(args, blob_callback=lint_non_binary_blobs)
-  filter.run()
+def main(argv1=None):
+  global is_relevant, lint_args, tmpdir, cat_file_process, filter
+
+  if isinstance(argv1, str):
+    argv1 = shlex.split(argv1)
```
newren, line 163: Where/how is this meant to be used?  It's a new usage that doesn't appear anywhere in current usage, so what's the motivating usecase(s) and what are you trying to accomplish?  Again, this may belong in a separate commit with its own explanation.

contrib/filter-repo-demos/git_lint_history.py
```diff
@@ -142,42 +151,47 @@ def lint_non_binary_blobs(blob, metadata):
...
     argv1 = shlex.split(argv1)
 
-  lint_args = parser.parse_args(argv1)
+  lint_args, rest = parser.parse_known_args(argv1)
   if not lint_args.command:
     raise SystemExit("Error: Need to specify a lint command")
-
+  if lint_args.relevant_glob:
+    if lint_args.relevant:
+      raise SystemExit("--relevant and --relevant-glob given both")
```
newren, line 170: This error message seems hard to understand.  Perhaps: "Error: --relevant and --relevant-glob are incompatible"?

contrib/filter-repo-demos/git_lint_history.py
```diff
@@ -142,42 +151,47 @@ def lint_non_binary_blobs(blob, metadata):
...
+    args.force = True
   if lint_args.filenames_important:
     tmpdir = tempfile.mkdtemp().encode()
     cat_file_process = subprocess.Popen(['git', 'cat-file', '--batch'],
                                         stdin = subprocess.PIPE,
                                         stdout = subprocess.PIPE)
-    filter = fr.RepoFilter(args, commit_callback=lint_with_real_filenames)
-    filter.run()
+    rfilter = fr.RepoFilter(args, commit_callback=lint_with_real_filenames)
+    rfilter.run()
```
newren, line 189: The rename from filter to rfilter appears to be gratuitous and unrelated.  Are you just trying to avoid a function defined elsewhere?  If so, that's probably a good change, but one that is not part of the functional changes being described and thus should really be part of its own commit.

contrib/filter-repo-demos/git_lint_history.py
```diff
@@ -91,13 +96,17 @@ def is_relevant(filename):
         help=("Python code for determining whether to apply linter to a "
               "given filename.  Implies --filenames-important.  See CALLBACK "
               "below."))
+parser.add_argument('--relevant-glob', metavar="GLOB", action='append',
+        help=("Glob of paths to include in filtered history. Can be specified "
+              "multiple times to get a union of paths."))
```
newren, line 101: The --relevant-glob and the forwarding of filter-repo options appear to be two separate things that probably belong in separate commits.

Also, the description talks about whether to include paths in history, but --path-glob already exists for that and since you're now forwarding git-filter-repo options it would seem to make sense to use it.  But is there a description error here?  Later, the code changes you make seem to suggest you are not using it to filter which blobs are included in history, but just using it to filter which blobs get linted while cleaning history while the others are left alone.

newren (changes requested):
Hi, thanks for your interest in filter-repo.  You have some interesting changes here, even if there's some small things to fix up here and there.  I've added a few comments, plus commits need to include a Signed-off-by trailer for inclusion.

newren:
No updates in 10 months so I'm going to close out.  Note that #316 is a separate implementation supporting just --refs.

PR #298, opened 2021-10-27
Improve Python Packaging

J-M0:
This PR updates and simplifies how git-filter-repo is built as a Python package. I added support for [PEP 517](https://www.python.org/dev/peps/pep-0517/) and [PEP 518](https://www.python.org/dev/peps/pep-0518/) and moved all of the configuration out of `setup.py` and into `setup.cfg` and `pyproject.toml`. Additionally, I removed some old and obsolete setuptools options.

Hopefully this solves some of the issues with installation via pip and by extension the issues with Git for Windows.


newren (changes requested):
Hi, thanks for your interest and contributions!  It's always nice to have someone else want to handle the PIP related stuff.

A few issues here:
  * All commits need to have a Signed-off-by trailer in order to be considered for inclusion.
  * Could you explain a bit more about why some of the code you are tossing is unneeded, especially since I don't know pip details but am trying to understand why different contributors have given different answers?
  * I really, really wanted to keep the root directory clean.  I'm quite annoyed that it has been polluted with so many files.  Everything beyond git-filter-repo and README.md represent some kind of very unfortunate tradeoff that makes me a bit unhappy.  As such, I am not enthusiastic about moving python packaging stuff moved to the root directory, especially since no end users should have to be mucking with it.  It seems like a perfect example of something that can and should be hidden in some subdirectory.  Are your subsequent changes still relevant if we drop the first commit, or do they all hinge on that one?

Makefile
```diff
@@ -126,15 +126,13 @@ github_release: update_docs
 
 pypi_release: # Has an implicit dependency on github_release because...
 	# Upload to PyPI, automatically picking tag created by github_release
-	cd release && python3 -m venv venv
-	cd release && venv/bin/pip3 install --upgrade setuptools pip
-	cd release && venv/bin/pip3 install twine wheel
-	cd release && venv/bin/python3 setup.py sdist bdist_wheel
+	python3 -m venv venv
+	venv/bin/pip3 install --upgrade pip build twine
+	venv/bin/pyproject-build
```
J-M0, line 131: [`build`](https://pypi.org/project/build/) is simple front end for building Python packages that implement PEP 517. This means that should you ever move from setuptools to a different build backend, this line doesn't have to change and will still work.

J-M0, line 131: [`build`](https://pypi.org/project/build/) is simple front end for building Python packages that implement PEP 517. This means that should you ever move from setuptools to a different build backend, this line doesn't have to change and will still work.

pyproject.toml
```diff
@@ -0,0 +1,5 @@
+[build-system]
+requires = ["setuptools>=51", "wheel", "setuptools_scm[toml]>=3.4"]
+build-backend = "setuptools.build_meta"
```
J-M0, line 3: This line is equivalent to the `use_scm_version` argument to `setup()`.

setup.cfg
```diff
@@ -16,17 +17,18 @@ classifiers =
...
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
     Programming Language :: Python :: 3
     Programming Language :: Python :: Implementation :: CPython
     Programming Language :: Python :: Implementation :: PyPy
 platforms = any
 license = MIT
 
 [options]
-scripts = git-filter-repo
```
J-M0, line 26: PyPA guidelines [recommend against](https://packaging.python.org/guides/distributing-packages-using-setuptools/#scripts) the `scripts` keyword. It's also redundant to specify the same name in `scripts` and `console_scripts` at the same time and would cause conflicts.

setup.cfg
```diff
@@ -16,17 +17,18 @@ classifiers =
...
 py_modules = git_filter_repo
 python_requires = >= 3.5
-setup_requires = setuptools_scm
```
J-M0, line 29: `setup_requires` is [deprecated in favor of](https://setuptools.pypa.io/en/latest/references/keywords.html) `pyproject.toml`

J-M0:
> All commits need to have a Signed-off-by trailer in order to be considered for inclusion.

Every single commit, or just the PR? This is the first time I've done something like that.

> Could you explain a bit more about why some of the code you are tossing is unneeded, especially since I don't know pip details but am trying to understand why different contributors have given different answers?

Sure. First, the reason why different contributors have given different answers is that Python packaging has been a mess for some time. It has however been getting better and more standards-based in [recent years](https://bernat.tech/posts/pep-517-518/). The general trend now is to move away from `setup.py` and running code and instead specify everything as statically as possible. Putting everything in `setup.cfg` and `pyproject.toml` moves towards that goal.

As for other changes I've made, I tried to make my commit messages explanatory. I made some other comments above (I meant for them to be below) explaining some individual lines' changes. If you still have questions after that, please let me know and I will do my best to explain them.

> I really, really wanted to keep the root directory clean...Are your subsequent changes still relevant if we drop the first commit, or do they all hinge on that one?

My motivation for moving the files to the root were two-fold. First was to make building the sdist consistent across platforms, particularly the fact that it doesn't work on Windows. Second, I was doing it with an eye to the future when setuptools gets support for [PEP 621] so that everything in `setup.cfg` can be moved to `pyproject.toml`. When that happens ther only extra file in the root directory would be `pyproject.toml`. I've also just pushed a commit which removes `setup.py` since it wasn't actually doing anything anyway for you the way I had it.

I could also try to make this future happen now by using flit instead of setuptools. That would come with the caveat though of you having to put a `__version__` attribute in `git-filter-repo` and remembering to bump that version number every release.

If you would really like to keep things in the `release` dir, I could try to rework my changes to make that happen.

newren:
> > All commits need to have a Signed-off-by trailer in order to be considered for inclusion.
> 
> Every single commit, or just the PR? This is the first time I've done something like that.

Every single commit.

> > Could you explain a bit more about why some of the code you are tossing is unneeded, especially since I don't know pip details but am trying to understand why different contributors have given different answers?
> 
> Sure. First, the reason why different contributors have given different answers is that Python packaging has been a mess for some time. It has however been getting better and more standards-based in [recent years](https://bernat.tech/posts/pep-517-518/). The general trend now is to move away from `setup.py` and running code and instead specify everything as statically as possible. Putting everything in `setup.cfg` and `pyproject.toml` moves towards that goal.

Ah, thanks for this background.

> As for other changes I've made, I tried to make my commit messages explanatory. I made some other comments above (I meant for them to be below) explaining some individual lines' changes. If you still have questions after that, please let me know and I will do my best to explain them.

Your commit messages might be self-explanatory to someone familar with PIP and setuptools, but to me they weren't.  Including some of the other comments you added in the commit messages would be helpful...particularly so that when the next contributor comes along who knows more about python packaging than me but less then you, they get some good pointers on why they shouldn't undo your work (and I possibly have a place to look to double check if I'm trying to guess which of you two is more up-to-date or more knowledgeable).
 
> > I really, really wanted to keep the root directory clean...Are your subsequent changes still relevant if we drop the first commit, or do they all hinge on that one?
> 
> My motivation for moving the files to the root were two-fold. First was to make building the sdist consistent across platforms, particularly the fact that it doesn't work on Windows. Second, I was doing it with an eye to the future when setuptools gets support for [PEP 621](https://www.python.org/dev/peps/pep-0621/) so that everything in `setup.cfg` can be moved to `pyproject.toml`. When that happens the only extra file in the root directory would be `pyproject.toml`. I've also just pushed a commit which removes `setup.py` since it wasn't actually doing anything anyway for you the way I had it.

Getting rid of setup.py, especially since there's no build process and it's just "hey, here's a single file script" sounds great.

> I could also try to make this future happen now by using flit instead of setuptools. That would come with the caveat though of you having to put a `__version__` attribute in `git-filter-repo` and remembering to bump that version number every release.

...and bump it in between releases so that someone cloning filter-repo and using it doesn't report the same version as an official release when the actual code is different.  Maybe that wouldn't be so bad given that development is much more bursty these days and much less active than when I was getting all the functionality in place.  But I'm still not sure if I want to make that switch.

> If you would really like to keep things in the `release` dir, I could try to rework my changes to make that happen.

That would be great, thanks!

J-M0:
Thanks for the feedback, I'll try to get working on it soon!

J-M0:
Hey sorry I went silent on this, I've been exploring some other avenues to get git-filter-repo on Windows and I've also been rethinking the changes I wanted to make to the python packaging. Since I have to basically redo all the commits to match the proper style, I'm just going to close this PR and open a new one if/when I'm ready with the new changes.

PR #299, opened 2021-10-29
Update mandb after installing the man page

erickeller:
when installing manually the manpage, one shall also update the mandb

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -118,7 +118,7 @@ So, installation might look something like the following:
 
    ```
    cp -a git-filter-repo $(git --exec-path)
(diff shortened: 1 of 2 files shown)


newren (changes requested):
Hi, thanks for the suggestion!  Could you add a Signed-off-by trailer to your commit message so I can include it?  While at it, could you change the second sentence/paragraph of the commit message to fix up the grammar a bit; perhaps:

"When manually installing the manpage, also update the mandb."

If you can do so, I'll be happy to include the change.

erickeller:
I did also add the trigger of mandb if the command exists in the Makefile

newren (approved):

newren:
Thanks!

PR #310, opened 2021-12-10
allow git-filter-repo clean-ignore args forwarding

fmigneault:
This allows the user to provide further argument parameters to `git-filter-repo` when running `clean-ignore`. A common example of when this is needed is when running `clean-ignore` on a repo not freshly cloned, causing below error. 
```
Aborting: Refusing to destructively overwrite repo history since
this does not look like a fresh clone.
```

Furthermore, I add the `git-clean-ignore` naming to allow git's auto-detection of the dash name to integrate it as sub-command automatically. 

With these change, it is possible to call as follows:
```shell
git clean-ignore --force
```

### DIFF
--- a/contrib/filter-repo-demos/clean-ignore
+++ b/contrib/filter-repo-demos/clean-ignore
@@ -18,6 +18,7 @@ near the top of git-filter-repo.
 import argparse
 import os
 import subprocess
+import sys
 try:
   import git_filter_repo as fr
 except ImportError:
@@ -65,7 +66,14 @@ class CheckIgnores:
     commit.file_changes = [x for x in commit.file_changes
                            if x.filename not in bad]
 
-checker = CheckIgnores()
-args = fr.FilteringOptions.default_options()
-filter = fr.RepoFilter(args, commit_callback=checker.skip_ignores)
-filter.run()
+
+def main():
+  checker = CheckIgnores()
+  args = fr.FilteringOptions.default_options()
+  filter = fr.RepoFilter(args, commit_callback=checker.skip_ignores)
+  filter.run()
+  
+
+if __name__ == '__main__':
+  main()
+
--- a/contrib/filter-repo-demos/clean-ignore
+++ b/contrib/filter-repo-demos/clean-ignore
@@ -69,7 +69,7 @@ class CheckIgnores:
 
 def main():
   checker = CheckIgnores()
-  args = fr.FilteringOptions.default_options()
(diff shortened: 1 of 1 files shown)


newren (changes requested):
Hi,

Sorry for being gone for so long.  Thanks for the interesting PR!

You actually have three logical changes here, which should probably be split into separate commits:
  * Calling fr.FilteringOptions.parse_args(sys.argv[1:]) instead of fr.FilteringOptions.default_options()
  * Adding a main() func and the check for __name__ == "__main__" stuff.
  * Making a symlink to clean-ignore named git-clean-ignore

I'm not sure I like the 3rd change; it would suggest you'd run it a lot, and rewriting history is not something you want to do lightly.

Also, the commit message needs a bit of a cleanup.  The standard style is supposed to be:

```
<area>: commit summary

Additional details motivating the change, providing history, etc. spread over a few paragraphs.

Signed-off-by: <Name and email>
```

Your commits are simple enough to allow you to omit the Additional details paragraph, but the sign-off is critical, and the <area> ("clean-ignore" in this case) is something I want included as well.

Anyway, if you restructure the commits and commit messages and drop the symlink, I'd be happy to merge it!  And sorry again for the long delay.  :-(

fmigneault:
Hi @newren 

The logic behind the symlink was to allow calling `git clean-ignore` since Git automatically handles the `git-<command>` naming convention. I am not against moving it to a separate PR (other commands could also use the same trick).

I can adjust the commits as requested. Is the "Signed-off-by:" part added manually or using some tool?

newren:
> The logic behind the symlink was to allow calling `git clean-ignore` since Git automatically handles the `git-<command>` naming convention. I am not against moving it to a separate PR (other commands could also use the same trick).

I'm totally aware of the `git-<command>` naming convention, but using it suggests that it's a command you'd want to use a lot so that you can use it as e.g. "git clean-ignore <args...>".  I think something is horribly wrong if you're using the tool a lot; it's a one-shot cleanup.  Also, I dislike generally squatting on lots of potential git-<command> names, because the existence of such commands have to be considered by Git developers whenever they add a new builtin command.  Therefore, I would like to only use one such name, for the high-level filter-repo command.

> I can adjust the commits as requested. Is the "Signed-off-by:" part added manually or using some tool?

Thanks.  Using the `--signoff` option to `git commit` should do it for you.  So, `git commit --amend --signoff` will add it to the existing commit you've got and let you adjust the commit message further to fix up the subject.  Then force push.

newren:
Thanks!

PR #326, opened 2022-01-17
INSTALL.md: correct a filename with mixed underscore and dash

marcows:


newren:
Sorry for the long delay, but thanks for the patch!  Merged!

PR #327, opened 2022-01-17
filter-repo: prepend the header line to the "ref-map" file

marcows:


newren:
Thanks!  And sorry for taking so long!

PR #330, opened 2022-01-24
Fix crash when filtering unrelated histories by refs

matejcik:
This is a sort of weird condition.

Consider repository `main-repo`, into which, at some point, another repository `side-repo` was merged via `--allow-unrelated-histories`. `main-repo` now has two root commits and two histories that converge at some point.
Next, pick a commit from `side-repo`'s own history, tag it with `side-tag` for convenience.

If you want to filter by `--refs=side-tag..HEAD`, for example, i.e., the selection starts inside `side-repo`'s history, then the output from `fast-export` is not sorted ideally. I forget the details right now, it made sense at the time ... in any case, the result is that an entry in `parents` can be an untranslated commit hash instead of ID.

So line 3461 will fail because `ID_TO_HASH[parent]` does not exist.

Fortunately we don't actually care because `get_file_changes` _wants_ a commit hash, so it's sufficient to use the raw entry.

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -3458,7 +3458,7 @@ class RepoFilter(object):
     # to the new first parent
     if parents and old_1st_parent != parents[0]:
       commit.file_changes = GitUtils.get_file_changes(self._repo_working_dir,
-                                                      ID_TO_HASH[parents[0]],
+                                                      ID_TO_HASH.get(parents[0], parents[0]),
                                                       commit.original_id)
       orig_file_changes = set(commit.file_changes)
       self._filter_files(commit)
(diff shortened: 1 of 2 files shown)


kdd0226:
二进制文件内 迁出项目后运行指定命令，重新提交结果。

newren:
Wow, nice find!  Thanks so much for digging this down and sending in a PR!

And I'm sorry for being so slow to respond.

In any event, I really need the Signed-off-by line in the commit message before I can merge.  If you can amend and add that, I'll be happy to merge it!

matejcik:
thanks for your response :) this works, I hope?  b07b17fcea556cd5b166bc920a0662f23817b9bd

newren:
Sad to lose this change, but can't accept it without the Signed-off-by.

PR #331, opened 2022-01-31
Makefile: add support for DESTDIR

Ikke:
Packaging `git-filter-repo` is a challenge, as the Makefile install target is mostly geared toward installing it on the local system. One reason is because the Makefile does not support `DESTDIR`. Another reason is that because it uses `cp`, all parent directories must already exist.

This pull request addresses both these issues.


newren:
Sorry for the long delay.  :-(

Anyway, changes look good to me so I've merged them.  Thanks for contributing!

PR #333, opened 2022-02-03
Fix path of git-filter-repo.(1|html)

mattiasdrp:
Since they are not at the root of the repo but in Documentation/(man1 | html) (the Makefile puts them here and the last releast tarball has them in it too)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -118,8 +118,8 @@ So, installation might look something like the following:
 
    ```
    cp -a git-filter-repo $(git --exec-path)
-   cp -a git-filter-repo.1 $(git --man-path)/man1 && mandb
-   cp -a git-filter-repo.html $(git --html-path)
+   cp -a Documentation/man1/git-filter-repo.1 $(git --man-path)/man1 && mandb
+   cp -a Documentation/html/git-filter-repo.html $(git --html-path)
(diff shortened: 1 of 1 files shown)


newren (changes requested):
Hi,

Sorry for taking so long to respond.

The changes make sense.  The commit message needs a little work -- lines should be wrapped at 72 characters, and I need a Signed-off-by line and a real email address to include the changes.  Could you update the commit to fix these?

mattiasdrp:
Hi,
No worries, we all have a lot of things to do :-)
I fixed my commit to include the changes
Thanks for this nice package

newren:
Hi,

Thanks for understanding, and I'm glad you've enjoyed the package.

Thanks for fixing the wrapping and adding a Signed-off-by.  It appears you've used a username rather than a real name, though.  Is that right?  As per https://github.com/git/git/blob/4fd6c5e44459e6444c2cd93383660134c95aabd1/Documentation/SubmittingPatches#L284-L286, I unfortunately need a real name for signoff.  (And the author and committer should match the Signed-off-by as well.)  Can you fix that?

mattiasdrp:
I'm not very fond of letting my real name everywhere on the Internet. I don't mind you taking the ownership of this PR and merging it with your name though so ;-)

PR #343, opened 2022-02-19
Correct mailmapping of empty email addresses

blepabyte:
Consider an example repo with 2 commits:
- Commit A by `User <some@email>`
- Commit B by `User <>` (the email field is left empty)

and running `git filter-repo --mailmap f` where `f` contains:
```
New user <new@email> <>
```

**Expected behaviour:** The author of commit B is updated to the new name and email address while commit A is unchanged. This is the behaviour observed when using a Git `.mailmap` file with the same contents. 

**Current behaviour:** Both commits A and B are modified to the new user and email. 

This happens because `not old_email` doesn't distinguish between `None` and an empty string (which is what gets matched by `<>`). This PR changes that to an `is None` check.

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -320,8 +320,8 @@ class MailmapInfo(object):
     for old, new in self.changes.items():
       old_name, old_email = old
       new_name, new_email = new
-      if (not old_email or email.lower() == old_email.lower()) and (
-          name  == old_name or not old_name):
+      if (old_email is None or email.lower() == old_email.lower()) and (
+          name == old_name or not old_name):
         return (new_name or name, new_email or email)
     return (name, email)
 


ryenus:
It would be great if the description is clarified with distinct sections for **What Happened** and **What's Expected**

newren (changes requested):
Hi,

Sorry for taking so long to reply.  :-(  Thanks for the report and the fix, though; very nice!

The code changes look good, but the commit needs a few fixes.  Can you use a real name and email address, and wrap the lines in the commit message at 72 characters?  If you can fix that up, I'd be happy to merge.

blepabyte:
Fixed up the commit details, should look better now.

newren:
> Fixed up the commit details, should look better now.

Thanks.  Could you verify that the name used is your real name?  (As per https://github.com/git/git/blob/3dcec76d9df911ed8321007b1d197c1a206dc164/Documentation/SubmittingPatches#L284-L286)?  It appears you may have used a last initial rather than name.

blepabyte:
One more try, with my full name this time.

newren:
Thanks, merged!

Issue #344, opened 2022-02-19
Test failure with git 2.35.0


stefanor:
From https://bugs.debian.org/1006056: We noticed that git-filter-repo's tests have regressed with git 2.35.0:

```
not ok 11 - check: empty -> more-empty-keepme using '--path keepme --prune-empty=always --prune-degenerate=always'
#	
#			# Clean up from previous run
#			git pack-refs --all &&
#			rm .git/packed-refs &&
#	
#			# Run the example
#			cat $DATA/$INPUT | git filter-repo --stdin --quiet --force --replace-refs delete-no-add "${REST[@]}" &&
#	
#			# Compare the resulting repo to expected value
#			git fast-export --use-done-feature --all >compare &&
#			test_cmp $DATA/$OUTPUT compare
#		
not ok 12 - check: empty -> less-empty-keepme using '--path keepme --prune-empty=never --prune-degenerate=never'
#	
#			# Clean up from previous run
#			git pack-refs --all &&
#			rm .git/packed-refs &&
#	
#			# Run the example
#			cat $DATA/$INPUT | git filter-repo --stdin --quiet --force --replace-refs delete-no-add "${REST[@]}" &&
#	
#			# Compare the resulting repo to expected value
#			git fast-export --use-done-feature --all >compare &&
#			test_cmp $DATA/$OUTPUT compare
#		
```

A bisect points to git/git@726a228dfb19fae8befa0f209436f8fae7919a58:

> fast-export: fix surprising behavior with --first-parent
>
> The revision traversal machinery typically processes and returns all children before any parent.  fast-export needs to operate in the reverse fashion, handling parents before any of their children in  order to build up the history starting from the root commit(s).  This would be a clear case where we could just use the revision traversal machinery's "reverse" option to achieve this desired affect.
>
> However, this wasn't what the code did.  It added its own array for queuing.  The obvious hand-rolled solution would be to just push all the commits into the array and then traverse afterwards, but it didn't quite do that either.  It instead attempted to process anything it could as soon as it could, and once it could, check whether it could process anything that had been queued.  As far as I can tell, this was an effort to save a little memory in the case of multiple root commits since it could process some commits before queueing all of them.  This involved some helper functions named `has_unshown_parent()` and `handle_tail()`.  For typical invocations of fast-export, this alternative essentially amounted to a hand-rolled method of reversing the commits -- it was a bunch of work to duplicate the revision traversal machinery's "reverse" option.
> 
[...]

PR #352, opened 2022-03-06
Update expected test data for git 2.35

stefanor:
Commit order from `fast-export --first-parent` has changed in git 2.35, see https://github.com/git/git/commit/726a228dfb19fae8befa0f209436f8fae7919a58

This will break the same tests on older git releases. I can't see an easy mechanism to skip these tests on older versions or carry multiple expected test outputs.

Fixes: #344


newren:
Sorry for the long delay.  Thanks a bunch for the fix!

PR #372, opened 2022-06-10
update replace-message - add support for optional commit ID filter

bairesearch:
update replace-message - add support for optional commit ID filter: end the line with '@commitID#' and a commit ID to restrict the replacement expression to a particular commit.

For example;
foo==>bar@commitID#d760d24cf74a0ee68134d4b83a43d1e8ea3b6e1c

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1858,7 +1858,9 @@ EXAMPLES
     messages.add_argument('--replace-message', metavar='EXPRESSIONS_FILE',
         help=_("A file with expressions that, if found in commit messages, "
                "will be replaced. This file uses the same syntax as "
-               "--replace-text."))
+               "--replace-text, with an optional commit ID filter. You can "
+               "end the line with '@commitID#' and a commit ID to "
+               "restrict the replacement expression to a particular commit."))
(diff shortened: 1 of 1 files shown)


Spiritaine (approved):

tenkirin:
@newren 
Any new progress about this one? I sought for this feature for a while and it turns out this feature is not merged yet.

newren:
I'd rather do this similar to the --file-info-callback added in commit 615720731589 (filter-repo: add a --file-info-callback, 2024-10-23), where the replace-text is available to the user in combination with both the file name and the file contents in a single callback, and the user can decide based on filename or file contents whether to apply the replace-text changes.  Granted, that's in the context of modifying files, whereas this is considering commit messages, but the idea still translates -- make a callback where the user has access to both the commit message, the commit original_id, and the replace-message rules and can choose whether to apply them.

rzr:
May I suggest to use this prefix by default then it can human/machine parsable

```
Origin:  {reference}
```

newren:
I appreciate the contribution, but with no response I think I'm going to close this one out based on my last comment.  Feel free to reopen if you modify to use that strategy, or just open a new PR with that strategy implemented.  Thanks!

PR #399, opened 2022-10-11
Makefile: abstract out the `install` binary used

### LINKED ISSUE #398: install -D not supported on macOS, maybe different on BSD
4ef1aabd4deafdfe7d559fdb2ba73f83477a52cc introduced the usage of `install -D`. The `-D` flag is not supported on macOS's install (which is simply very old), and on BSD it seems that it might not have the same meaning. Compare GNU install:

       -D     create all leading components of DEST except the last, or
              all components of --target-directory, then copy SOURCE to
              DEST

https://man7.org/linux/man-pages/man1/install.1.html

To BSD install:

      -D	destdir
	     Specify the DESTDIR (top of the file hierarchy) that the items
	     are installed in to.  If -M metalog is in use, a leading string
	     of	"destdir" will be removed from the file	names logged to	the
	     metalog.  This option does	not affect where the actual files are
	     installed.

https://www.freebsd.org/cgi/man.cgi?query=install

If removing `-D` is too cumbersome, to make packagers' lives easier (I maintain the [MacPorts](https://www.macports.org/) port of git-filter-repo) it would be nice to abstract the choice of install binary into a variable that could be overridden: `INSTALL = install`; then we can substitute GNU install e.g. by doing `make INSTALL=/opt/local/bin/ginstall`.

amake:
Fixes #398

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # A bunch of installation-related paths people can override on the command line
 DESTDIR = /
+INSTALL = install
 prefix = $(HOME)
 bindir = $(prefix)/libexec/git-core
 localedir = $(prefix)/share/locale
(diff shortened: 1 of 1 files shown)


newren:
Change looks good, but the commit message needs to be self-standing.  Could you change the "Fixes #398" line to a description that doesn't reference GitHub issues?  Something like
```
The -D option to `install` is not portable.  Abstract out the `install`
binary used so that other platforms can use e.g. ginstall to circumvent
this problem.
```
I'll be happy to merge once the Fixes line is fixed.

amake:
Done

PR #405, opened 2022-10-23
Documentation/windows installation

RaduTerec:
Added installation instructions for Windows.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,6 +6,7 @@
   * [Detailed installation explanation for
      packagers](#detailed-installation-explanation-for-packagers)
   * [Installation via Makefile](#installation-via-makefile)
+  * [Installation on Windows](#installation-on-windows)
   * [Notes for Windows Users](#notes-for-windows-users)
 
 # Pre-requisites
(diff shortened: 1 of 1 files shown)


newren:
Thanks, but why did you think you should install it with a `.py` ending?  git-filter-repo should be recorded exactly as its name says: as `git-filter-repo`.  Not as `git-filter-repo.py`.

Anyway, there is already a windows-specific section, and your first step duplicates the first item mentioned in it.  Your steps 2-4 merely duplicate the generic instructions, other than the error of renaming the file to `git-filter-repo.py`, so I'm curious what led you down that road.

RaduTerec:
Hi!

Thanks for taking the time to reply. When I downloaded the file, it had the
extension "txt", so I assumed that can't be right. Since I needed python to
run it, I assumed the right extension would be ".py". However, I just
checked it and it works fine without any extension.I will change this part
in my addition.

I went through the Windows section and thought I could improve it by
providing a more concise description of installing the script with a
fitting title (I could add this as a subsection of the Windows issue
section), a direct link to Python3 in Windows Store instead of a link to
the MS page, where you have another link to the Store page and by saying
that it mustn't have any extension.

My goal is to help the community, so if there is something I should change
or rephrase to make it fit better within the document, I'll be happy to do
it :)

Thanks for creating and mainting this repo!

Radu

---
The piano keys are black and white, but they sound like a million colors in
your mind.


Am Mo., 24. Okt. 2022 um 17:05 Uhr schrieb Elijah Newren <
***@***.***>:

newren:
Hi,

The note that Windows stupidly suggests an unrelated extension is good information; thanks for reporting it.  It does sound like something we should address.

In your current pull request, the instructions are still mostly just awkwardly duplicating stuff elsewhere in the document; I don't see the need for a new section.  Instead, let's look at the two things of note you added:
  * The comment about making sure to not save the file under an alternate name could be added to the existing text that talks about downloading the git-filter-repo file.
  * The more direct python link could potentially be added to the other location as well, except that the link appears to be specific to python-3.10, meaning I'd have to update the link with every new python release, which I'm not willing to do.

Additionally, the commits should be clean and each one has to have the appropriate Signed-off-by.  "fixup" commits make history ugly, and should just be squashed into the commit they are fixing up.

newren:
Hey, not sure if I lost you.  I tried putting together an example of what I had in mind over at https://github.com/newren/git-filter-repo/commit/436b83ea1433b4ca754f4f96990f5d2c6d73dcc0; it incorporates the improvements you suggest (people can correctly install by just looking at the pre-reqs and simple instructions sections, closer to a direct link to the python needed, warning about Windows stupidly adding an unwanted extension when downloading a file) and does so without duplicating the instructions.  Let me know if that still causes confusion.

RaduTerec:
Thanks for stitching this commit together. The phrasing looks good to me, especially now that you mentioned about the file extension and the link to download python.

Issue #413, opened 2022-11-30
_parse_user aborts if line does not match expected regex


larsks:
I was trying to remove all the release archives from a clone of http://smarden.org/git/runit.git with the following command:

```
git-filter-repo --invert-paths --path-glob '*.tar.gz' --force
```

But that fails with:

```
Traceback (most recent call last):
  File "/home/lars/bin/git-filter-repo", line 4006, in <module>
    main()
  File "/home/lars/bin/git-filter-repo", line 4003, in main
    filter.run()
  File "/home/lars/bin/git-filter-repo", line 3938, in run
    self._parser.run(self._input, self._output)
  File "/home/lars/bin/git-filter-repo", line 1412, in run
    self._parse_tag()
  File "/home/lars/bin/git-filter-repo", line 1291, in _parse_tag
    (tagger_name, tagger_email, tagger_date) = self._parse_user(b'tagger')
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lars/bin/git-filter-repo", line 1078, in _parse_user
    (name, email, when) = user_regex.match(self._currentline).groups()
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'groups'
fatal: stream ends early
fast-import: dumping crash report to .git/fast_import_crash_1855502
```

Modifying the code to add a `breakpoint()` just before that line, we see:

```
(Pdb) p self._currentline
b'tagger Gerrit Pape <pape@smarden.org>\n'
```

Whereas `user_regex` is looking for:

```
(Pdb) p user_regex
re.compile(b'tagger (.*?) <(.*?)> (.*)\n$')
```

Rewriting the regex to match is trivial, but it wouldn't provide a value for `when`, so I'm not sure of the correct fix.

newren:
Thanks for the report.

I'm not sure I want filter-repo to try to handle all forms of repository corruption, and a tag without a tagger date is malformed.  You can verify by just running `git fsck` on the repo.  I manually created a repo with such a problem (with a tag named `1.0.0`), and it reports as follows:
```
$ git fsck
error in tag b8c23fecae8bb7925ef4fb18203872adb4a0727f: missingSpaceBeforeDate: invalid author/committer line - missing space before date
Checking object directories: 100% (256/256), done.
Checking objects: 100% (3/3), done.
```

I can manually fix this repository corruption as follows:
```
[...]

PR #414, opened 2022-11-30
Handle tags with no date without blowing up

larsks:
In some situations, git-fast-export generates tagger lines with no date, such as:

    tag v1_7_1
    mark :758
    from :592
    original-oid 52e4c981c5ee8c4225316b418775e95dd436b6ef
    tagger Gerrit Pape <pape@smarden.org>
    data 0

This would cause git-filter-repo to fail because the regular expression used to parse user information wouldn't match.

This commit modifies the regular expression so that it user information without a date, and then emit a date of "0 +0000" when writing out tags that have no date.

I'm not sure this is an ideal solution; I think a better solution would be for the tag to inherit the date from the commit to which it points, but that seemed a little more involved and this at least lets filter-repo successfully complete.

Addresses #413

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -767,7 +767,10 @@ class Tag(_GitElementWithId):
     self.tagger_email = tagger_email
 
     # Store the date
-    self.tagger_date  = tagger_date
+    if tagger_date:
+        self.tagger_date  = tagger_date
+    else:
+        self.tagger_date = b"0 +0000"
 
     # Store the tag message
     self.message = tag_msg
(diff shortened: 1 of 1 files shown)


Cobe13 (approved):

newren:
Thanks for digging and sending in the PR!  The code looks clean and well done...but I'm not sure I want filter-repo to start trying to handle fixing all forms of repository corruption, and a tagger without a date is corrupted.  I left more comments about this in #413.

newren:
Closing as per comment in #413

PR #432, opened 2023-02-08
filter-repo: add missing "will" to generated README

Hattes:
I considered "may" or "can", but "will" makes more sense to me in context.


Hattes:
I was caught by the "Signed-off-by" check. Following the instructions to change my commits didn't work. Quite frustrating. I will give up on this.

PR #443, opened 2023-03-27
README: fix typo to clarify design rationale

noobsauce101:
While I am not a big fan of PRs to to fix typos, the README.md `Design Rationale` section has a small typo ("any" instead of "each") that confused me on my first read-through so I added a tiny PR to fix it, since it hinders understanding.

I am assuming that the design rationale was that there were no existing tools which implemented all of the first 8 traits, and none that implemented even a single one of the final four traits. This change just makes that clear (if that was the intended interpretation!).

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,9 +308,9 @@ But this comes with some nasty caveats and limitations:
 # Design rationale behind filter-repo
 
 None of the existing repository filtering tools did what I wanted;
-they all came up short for my needs.  No tool provided any of the
-first eight traits below I wanted, and all failed to provide at least
-one of the last four traits as well:
+they all came up short for my needs. No tool provided any of the
+first eight traits below I wanted, and no tool provided more than
+two of the last four traits either:
 
   1. [Starting report] Provide user an analysis of their repo to help
      them get started on what to prune or rename, instead of expecting


newren:
> I am assuming that the design rationale was that there were no existing tools which implemented all of the first 8 traits, 

Why would you assume that existing tools implemented some of the first 8 traits?  Are there particular ones you think other systems implemented?

As best I can tell, git-filter-repo provided none of the traits I listed in 1-12, bfg-repo-cleaner only provided 9 & 12 (some might think it provides 6, but look more closely at the comment at the top of https://github.com/newren/git-filter-repo/blob/main/contrib/filter-repo-demos/bfg-ish and you'll notice it fails there, and not just due to bugs but because of design limitations), and direct use of fast-export | fast-import only provides item 12.

So, the original wording is best -- no tool provided any of the first 8 attributes.

> and none that implemented even a single one of the final four traits.

No, as mentioned above, bfg-repo-cleaner provided two of the final four traits, and manually tweaking a fast-export stream before feeding to fast-import provided one of the final four traits.

noobsauce101:
> Why would you assume that existing tools implemented some of the first 8 traits? Are there particular ones you think other systems implemented?

I wasnt assuming anything about any the existing implementations, I was assuming what you had meant by the sentence "No tool provided any of the first eight traits".

> So, the original wording is best -- no tool provided any of the first 8 attributes.

Okay but then why would the readme say "No tool provided any of the first eight traits below I wanted, and all failed to provide at least one of the last four traits as well"  - isnt that the same as saying "No tool provided any of the twelve traits below"?

If no tool provides any of the first 8, and all of them fail to provide at least one of the last 4, then none of them provide any of the 12 traits.

> As best I can tell, git-filter-repo provided none of the traits I listed in 1-12, bfg-repo-cleaner only provided 9 & 12 (some might think it provides 6, but look more closely at the comment at the top of https://github.com/newren/git-filter-repo/blob/main/contrib/filter-repo-demos/bfg-ish and you'll notice it fails there, and not just due to bugs but because of design limitations), and direct use of fast-export | fast-import only provides item 12.

So maybe it should be reworded to "No tool provided any of the first eight traits below I wanted, and all failed to provide `more than two` of the last four traits as well" ?

newren:
> > So, the original wording is best -- no tool provided any of the first 8 attributes.
> 
> Okay but then why would the readme say "No tool provided any of the first eight traits below I wanted, and all failed to provide at least one of the last four traits as well" - isnt that the same as saying "No tool provided any of the twelve traits below"?

Ah, so the second half of the sentence is ambiguous (which I didn't previously recognize), and your misunderstanding of what I meant by the second half made you doubt the first half of the statement instead of the second half?

Let's say, for example, that one tool failed to provide item 9, one failed to provide item 10, and one failed to provide item 12.  The second half of the statement in that case could have been "and all failed to provide one of the last four traits."

Now, let's modify that hypothetical slightly and say that one of the alternatives, instead of providing 3 of the 4 wanted traits only provided 1 of the 4.  Meaning, two tools failed to provide one trait, and one tool failed to provide three traits.  We could then change the second half of the statement in that case to "and all failed to provide at least one of the last four traits."

> If no tool provides any of the first 8, and all of them fail to provide at least one of the last 4, then none of them provide any of the 12 traits.

No, not necessarily.  There are two ways to interpret "all of them fail to provide at least one", and you and I picked opposite possibilities without realizing there was another interpretation of that statement.  It's just inherently ambiguous, so we should fix it.

> > As best I can tell, git-filter-repo provided none of the traits I listed in 1-12, bfg-repo-cleaner only provided 9 & 12 (some might think it provides 6, but look more closely at the comment at the top of https://github.com/newren/git-filter-repo/blob/main/contrib/filter-repo-demos/bfg-ish and you'll notice it fails there, and not just due to bugs but because of design limitations), and direct use of fast-export | fast-import only provides item 12.
> 
> So maybe it should be reworded to "No tool provided any of the first eight traits below I wanted, and all failed to provide `more than two` of the last four traits as well" ? (or `at least two` since you mentioned that bfg-repo-cleaner manages to provide 2 of the last 4)

How about "No tool provided any of the first eight traits below I wanted, and no tool provided more than two of the last four traits either." ?

noobsauce101:
Yes- exactly! Thats what I had been confused by. Sure that sounds good, ill change it

README.md
```diff
@@ -308,9 +308,7 @@ But this comes with some nasty caveats and limitations:
 # Design rationale behind filter-repo
 
 None of the existing repository filtering tools did what I wanted;
-they all came up short for my needs.  No tool provided any of the
-first eight traits below I wanted, and all failed to provide at least
-one of the last four traits as well:
+they all came up short for my needs. No tool provided any of the first eight traits below I wanted, and no tool provided more than two of the last four traits either:
```
newren, line 311: Could you wrap this long line like the rest of the document?  After that, I'll be happy to apply.

noobsauce101: Ah yeah sorry about that -done now

newren:
Thanks!

Issue #249, opened 2021-05-03
Setup.py throws on windows unless running as administrator


CervEdin:
This isn't exactly a severe issue but running `setup.py` on windows throws unless running as an administrator since creating a symlink on Windows requires administrator privileges'.

Possible solutions:

- Add documenation to run as administrator
- Change the symlink to hardlink on Windows

Not sure exactly sure what the symlink is for, so I'm not sure what the best fix is.

Thanks for an awesome script, I love it.

newren:
I'm glad you are enjoying the script.

setup.py exists solely for uploading git-filter-repo to PyPI, which only someone doing releases should run.  Someone else contributed it (and I have to admit I don't really understand all the setup and pypi stuff).  I presume copies would work as well as symlinks.

Out of curiosity, what are you trying to use setup.py for?  It's not referenced in the installation documentation anywhere.

CervEdin:
Actually I think I just assumed this was the reason the install failed when running a non-elevated shell. I believe I was using scoop.

newren:
setup.py was removed in 2c9cc90609a7 (Modernize and simplify Python packaging, 2023-09-24), making this obsolete.  I suspect scoop has long since fixed up their installation bug, but if not, someone can file an issue with them.

Issue #307, opened 2021-12-04
Symbolic link not working on Windows


peterleif:
These 2 files are commited as symbolic links

- git_filter_repo.py (https://github.com/newren/git-filter-repo/blob/main/git_filter_repo.py)
- filter-branch-ish (https://github.com/newren/git-filter-repo/blob/main/contrib/filter-repo-demos/filter-branch-ish)

However, these symbolic links are not treated as symbolic links on Windows (they are substituted by text files instead).

Could you consider stop commiting these files as symbolic links?

Thanks

CervEdin:
you can enable symlinks in windows

newren:
If I make them be copies of the file in question, then I have to update them both whenever I update the main file.  That makes history rather ugly.  If you don't want to enable symlinks on Windows, just make them be copies yourself and they'll function just fine.

### Additional Background

I would have preferred to not have git-filter-repo.py exist at all, but python made importing files that don't end in .py really painful and makes all the contrib examples super ugly.  But I could just use git-filter-repo.py because `git X` will work and call `git-X` without the extension.  So, I need both to exist, but I don't want to update both.  Symlinks are a super nice solution, but others can switch it to a copy locally if they want.

filter-branch-ish only exists so that people can find "filter-branch" and how to replace it.  The really name of the script is "filter-lamely" because filter-branch is a horrendous design that is going to cause people problems and I want the name of the script to warn people against its use.  They symlink also makes it clear that the file has a real name elsewhere.

So, both serve good purposes, and I'll keep both as symlinks.  You can turn them into copies on your end if you please; nothing relies on them remaining symlinks.

PR #455, opened 2023-04-14
Remove symlink that breaks install on Windows

rdbisme:
e.g. Using scoop

See: https://github.com/ScoopInstaller/Scoop/issues/4647:

Fix: #249

### DIFF
--- a/git_filter_repo.py
+++ b/git_filter_repo.py
@@ -1,1 +1,0 @@
-git-filter-repo


rdbisme:
Just removing the symlink to the python file should be good, because when publishing to PyPI, it gets created by the `setup.py`

newren:
We should not rely on installation packages to add a symlink; it should be there in the original clone.  Not to mention that I'm think that pip support and setup.py was a mistake, even if a well-intentioned one.  Anyway, since you are relying on an installation package (namely from scoop) to get the package, they should be responsible to provide a working installation, including replacing the symlink with a copy.  Feel free to file a bug with them.

See also issue #307

rdbisme:
Hello, thanks for your answer. The problem is that on unprivileged Windows, for what I can understand, uncomoressing archives with symlinks is impossible without admin rights. That would require sourcing from the git repository instead of the release tarball. 

What about creating the symlinks at release time, and having a special package for windows where symlinks are replaced by copies?

That won't require for you to maintain the copies, they will just be created at release time?

Your tarball is effectively not cross platform at the moment because of those symlinks.

PR #464, opened 2023-05-08
feat: demo script to proc and replace files with shell commands

ankostis:
Adapted  from `lint-histoty.py` demo,  but command's output can be a different file which is added in the history,
and optionally replaces completely the source file.

For context, I needed this one to post-process my jupyterlab notebooks (json-files) and convert them into `py:percent` regular python files with [jupytext](https://jupytext.readthedocs.io/) so as git-diff to work decently.

### DIFF
--- a/contrib/filter-repo-demos/README.md
+++ b/contrib/filter-repo-demos/README.md
@@ -17,6 +17,7 @@ clean-ignore         |Delete files from history which match current gitignore ru
 filter-lamely (or filter&#8209;branch&#8209;ish) |A nearly bug compatible re-implementation of filter-branch (the git testsuite passes using it instead of filter-branch), with some performance tricks to make it several times faster (though it's still glacially slow compared to filter-repo).
 bfg-ish              |A re-implementation of most of BFG Repo Cleaner, with new features and bug fixes.
 convert-svnexternals |Insert Git submodules according to SVN externals.
+proc-history         | Run shell command on each file in history and add or replace the file with cmd's output.                                                                                                                               
 
 ## Purpose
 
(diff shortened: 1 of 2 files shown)


ankostis:
Is there something i can do to approve CI workflow for this PR?

huyz:
Could there be a way to add this functionality to the original `lint-history` with a flag?

newren:
Sorry for the long delay.

This PR is very interesting.  I think @huyz's comment is a good one, though: this would give us a very heavy amount of code duplication (and even a copy of a now-obsolete large comment block) between lint-history and your script.  While you make several stylistic changes throughout, those stylistic changes are equally applicable to lint-history and we really should probably either have both take the new style or both take the old style (to be determined individually for each stylistic change).  From a high level point of view, though, namely from the user's perspective, your script really is just lint-history with two additional parameters (--outfpath && --drop-src).  As such, I think it'd be better to structure it as an extension to that file.

Also, the commit message looks like it got mangled.

But, at an even higher level, why is this script needed?  Users could already run e.g.
```
lint-history --relevant 'return filename.lower().endswith(b".ipynb")' -- jupytext --to 'py:percent'
git filter-repo --filename-callback 'return filename[:-6] + b".py" if filename.lower().endswith(b".ipynb") else filename'
```
In other words, modify the contents by calling lint-history, and then rename in a separate step (using git filter-repo).

DEMO861 (approved):

newren:
I added a --file-info-callback as part of commit 615720731 (filter-repo: add a --file-info-callback, 2024-10-23). That commit adds a callback which allows you to work with both the filename and file contents (and file mode).  That callback essentially obsoletes both the lint-history script and this one; since you have access to both the filename and the file contents, you can choose to modify only certain files and simultaneously rename them, for example.  So, I'll close this PR out as no longer being relevant.  Thanks for your interest and contributing, though; this request along with about half a dozen others led me to step back and create the new more general callback to enable folks to handle all the various different cases.

PR #470, opened 2023-05-22
Minor typos

zsobhani:
Fixed a few minor typos. 

'emails addresses' >> 'email addresses'
'reset' >> 'alias' in Alias class

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1933,7 +1933,7 @@ EXAMPLES
         help=_("Python code body for processing names of people; see "
                "CALLBACKS section below."))
     callback.add_argument('--email-callback', metavar="FUNCTION_BODY_OR_FILE",
-        help=_("Python code body for processing emails addresses; see "
+        help=_("Python code body for processing email addresses; see "
                "CALLBACKS section below."))
     callback.add_argument('--refname-callback', metavar="FUNCTION_BODY_OR_FILE",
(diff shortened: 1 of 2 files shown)


git-filter-repo
```diff
@@ -873,15 +873,15 @@ class Alias(_GitElement):
 
   def __init__(self, ref, to_ref):
     _GitElement.__init__(self)
-    # Denote that this is a reset
+    # Denote that this is a alias
```
marcows, line 876: an alias

zsobhani: @marcows Thanks for that catch! LOL, I'll trade you a typo for a typo. I was also reviewing more closely branch names from contributors, which convention I did not follow. I am going to open a new pull request from a different branch. Thanks for your patience!

marcows: The branch naming like zs/xxx is used by Elijah locally (visible to us because of local merge without the GitHub merge button), it is not a convention for you to follow. And even if it was, you just should have updated your branch instead of creating a new PR.

zsobhani: @marcows Thanks for taking the time to correct me here. I'll see if I can do it right. I applied your correction in the original branch, but perhaps made a further mistake by doing an interactive rebase to squash the fix-up with the original commit attempting to fix the "alias" typo. I hope that wasn't another no-no.

PR #477, opened 2023-06-03
Add documentation generation

jpoiret:
This uses the manpage XSL and asciidoc config files from git 2.40.1 to build the documentation on its own. This work was done to package this in GNU Guix, inspired by the similar Debian patches [1]. This of course doesn't touch any of the maintainer stuff, but it could be transitioned to use this instead.

[1] https://salsa.debian.org/python-team/packages/git-filter-repo/-/tree/debian/latest/debian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /Documentation/html/
 /Documentation/man1/
+/Documentation/manpage-base-url.xsl
 /t/test-results
 /t/trash directory*
 /__pycache__/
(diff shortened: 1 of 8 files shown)


newren:
Sigh...this means yet another thing we borrow from git that will continue going further and further out of date...though it probably makes sense since those things don't get changed that often and being able to generate our own docs is nice.

It feels a little weird that when I cd into the Documentation directory and run a plain 'make', that it builds neither the man or html targets and instead runs something else.  Could you fix that?

Most importantly, the commit is missing your Signed-off-by line (https://github.com/git/git/blob/master/Documentation/SubmittingPatches#L351), which is needed to accept the contribution.  Can you add it?

Other than those two minor issues, the patch looks good.

DEMO861 (approved):

newren:
No response to the request to fix the missing Signed-off-by.  Sad to lose these changes, but I can't merge without that.  Closing.

PR #479, opened 2023-06-10
fix manual htmplreview link

deemp:
Seems like htmpreview can't reach files from /blob/. Tried with Firefox.


deemp:
Upd: hmm, strangely, now, it does. So, the original link is fine.

Unclelimzt (commented):
https://github.com/newren/git-filter-repo/pull/479#issue-1750916855

PR #485, opened 2023-06-29
fix: TypeError: environment can only contain strings

rasa:


newren:
Could you talk more about where you hit the error that spurred the two commits in this patch?  What kind of filtering were you doing that you were able to trigger an error here?  Is it an operating system specific change?  Does it limit filter-repo's capabilities on other platforms?  Why are you specifically limiting to ascii rather than some other subset such as utf-8?

Your commit messages also need your signoff (https://github.com/git/git/blob/master/Documentation/SubmittingPatches#L351) if they are to be included.

rasa:
> Could you talk more about where you hit the error that spurred the two commits in this patch? What kind of filtering were you doing that you were able to trigger an error here? 

It was a while ago, so I don't recall exactly.

> Is it an operating system specific change? 

The failure was on Windows.

> Does it limit filter-repo's capabilities on other platforms? 

Didn't test.

> Why are you specifically limiting to ascii rather than some other subset such as utf-8?

The patch was to deal with the fact that something in my `os.environ` was causing the error. I don't recall what it was, and that environment doesn't exist any more.

Sorry I can't be more helpful. I will close this PR, and if I encounter the issue again, I will describe the exact details that led to the patch.

PR #498, opened 2023-08-12
fix duplicate setuptools entry point script

tjni:
There is a subtle issue when installing from the wheel that's built. By declaring a console_script entry point, when the wheel is installed, a script named `git-filter-repo` is created. At the same time, inside setup.cfg, a script named `git-filter-repo` is declared (and a symlink is created in the release directory). This causes a collision when installing the wheel.

When the wheel is installed using `pip`, the collision is ignored. When the wheel is installed using [installer](https://installer.pypa.io/en/stable/), an error is raised. This PR fixes the issue by removing the script and relying only on the console_script entry point.

### DIFF
--- a/release/setup.cfg
+++ b/release/setup.cfg
@@ -23,10 +23,13 @@ platforms = any
 license = MIT
 
 [options]
-scripts = git-filter-repo
 py_modules = git_filter_repo
 python_requires = >= 3.5
 setup_requires = setuptools_scm
+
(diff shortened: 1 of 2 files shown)


newren:
Since this conflicts with #507, and I like how that one nukes the release/ directory, I have merged it and I'm going to close this one.  Thanks for the patch, though!

PR #507, opened 2023-09-29
Improve Python packaging, take 2

J-M0:
Hi, it's been a while but I'm back again to try to make the python packaging a little better.

I said [previously](https://github.com/newren/git-filter-repo/pull/298#issuecomment-965938762) that setuptools was going to do away with `setup.py` and have a new way to statically specify package data. That future is now. All the package information is now specified in `pyproject.toml` and `setup.py` is no more.

With the release directory removed and `pyproject.toml` added, there's a net zero new files polluting the root directory. As an added bonus, Python tools are standardizing around `pyproject.toml` to store their configuration in. That means if you want to introduce other QA tools to the dev process, there won't be a ton of config files littering the root.

Supersedes #298, #498.


newren:
Cool, I'll give it a shot.

PR #511, opened 2023-10-10
Allow clean-ignore to run without arguments.

CEBasile:
clean-ignore is written to allow leverage of filter-repo to run additional operations on top of clean-ignore. If no arguments are specified nothing is passed to filter-repo and it errors out as it's expecting arguments. I chose "--force" based on previous comments from newren.

### DIFF
--- a/contrib/filter-repo-demos/clean-ignore
+++ b/contrib/filter-repo-demos/clean-ignore
@@ -69,7 +69,10 @@ class CheckIgnores:
 
 def main():
   checker = CheckIgnores()
-  args = fr.FilteringOptions.parse_args(sys.argv[1:])
+  if sys.argsv[1:]:
+    args = fr.FilteringOptions.parse_args(sys.argv[1:])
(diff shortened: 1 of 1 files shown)


newren:
Oh man, I'm such a bad maintainer.  I apologize for taking so long to get back to you.

While investigating the issue behind 2800bcc1007e (clean-ignore: support utf-8 filenames found in .gitignore, 2024-07-02), I noticed the same issue and fixed it in a way compatible with how it was fixed for lint-history; see 5c54a5332b8e (clean-ignore: do not require additional arguments to be passed, 2024-07-02).

So, it's now fixed but differently.  Sorry for not getting back to you sooner and suggesting you make the tweak.  For future reference, it'd also be better to squash your commits into one (since these are not independent changes and the second just looks like a fixup of the first), and to wrap commit messages at 72 characters.

Anyway, I appreciate the contribution and feel bad for stepping on your toes.

PR #528, opened 2023-12-13
args: allow to use multiple `--refs` arguments

nslsrv:
Refs arguments are passed to `git rev-list` allowing user to configure the set of commits to operate on. IMHO it's important to be able to pass through any desirable rev-list configurations. This often requires using multiple `--refs` arguments.

Examples:
* `git filter-repo --refs head1 --refs head2`
* `git filter-repo --refs='--glob:refs/heads/*' --refs='--glob:refs/tags/*'`
* `git filter-repo --refs='--glob:refs/heads/*' --refs='--not' --refs='master`

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1990,7 +1990,7 @@ EXAMPLES
     #     commit has an excluded ancestor as a parent we have no way of
     #     knowing what it is an ancestor of without doing a special
     #     full-graph walk.
-    misc.add_argument('--refs', nargs='+',
+    misc.add_argument('--refs', action='append',
         help=_("Limit history rewriting to the specified refs.  Implies "
                "--partial.  In addition to the normal caveats of --partial "
                "(mixing old and new history, no automatic remapping of "


nslsrv:
@newren Hi, could you please review this one when you have some free time? This patch is good for cases when you need to rewrite ref history partially.

newren:
Sorry for the delay.

The PR is interesting, but it is going to break people that have already been passing multiple refs like so:
```
   git filter-repo --refs head1 head2
```
or
```
   git filter-repo --refs ^master head1 head2
```

This would also invalidate part of the documentation, and would cause one of the tests in the regression testsuite to fail.  This might make sense as an alternative way to have done things from the beginning, but I don't like breaking existing uses.

newren:
I'm going to have to close this one out because of the backward compatibility issues mentioned above.  This may well have made sense as a better way to do things from the beginning, but we can't go back in time.

Thanks for sending it in, though!

PR #531, opened 2023-12-22
Add --no-gc option

vapaamies:
Feature to skip `git gc` run, useful for multiple `git-filter-repo` runs.


newren:
The --no-gc option seems reasonable; could you drop the pycharm commit, include the rationale from this PR in your commit message, and add your signoff (https://github.com/git/git/blob/master/Documentation/SubmittingPatches#L351)?

Issue #76, opened 2020-04-15
Wrong behavior of 'clean-ignore' contrib


leni8ec:
I have a '.gitignore' file with this rule:
`!/[Aa]ssets/**/*.meta`

And when i execute this command:
` py ..\git-filter-repo\contrib\filter-repo-demos\clean-ignore`
all `.meta` files are **excluded** from the repository, although this should not be.

After removing this rule from `.gitignore` - it works correctly (`.meta` files remain in the repo).

>  Git 2.26.0.windows.1, Windows 10, PowerShell 7

newren:
The clean-ignore contrib script doesn't directly read .gitignore files at all; it has no notion of the correctly rules for what is ignored or not.  It defers all such checking to a call to `git check-ignore --stdin`.  And if that's wrong, then the bug is in git core rather than in git-filter-repo.  So the real question, is what does that report for you? 

Can you create a new git repo (`git init newrepo && cd newrepo`), then putting just the line into .gitignore, then try running `echo "assets/somefile.meta" | git check-ignore --stdin` and report what output you see?

If I do that with git-2.24.0, I see "assets/somefile.meta" (plus a trailing newline), because with that rule, git-2.24.0's check-ignore command thinks that the file IS ignored.

If I do that with git-2.26.0, I get no output, because I fixed that annoying bug in git check-ignore in the 2.26.0 cycle making sure that check-ignore understands the actual ignore rules (see https://git.kernel.org/pub/scm/git/git.git/commit/?id=7ec8125fba96f47b00bb0cab3ed716557e81e7e6).

I know you reported Git 2.26.0.windows.1, which is the bigger surprise to me here.  Is there some way that you also have an earlier git installed, and git-filter-repo is picking that one up?  Do you have other rules in your .gitignore file that are overruling this one?

In order to rule out other lines, it might be useful to run
`echo "assets/${SOMEFILE}.meta" | git check-ignore --stdin --verbose` for various values of ${SOMEFILE} to see which gitignore line(s) they are matching.

newren:
Any update?

newren:
No response; I'll have to assume that git-filter-repo found an older git and invoked it, giving you the bad old check-ignore behavior.  Feel free to reopen if that's not the case and you can answer the questions above.

jmgate:
I'm seeing what I think is the same issue.  My repository's `.gitignore` has, among other things,
```.gitignore
*.mat
!path/to/particular.mat
```
[...]

PR #560, opened 2024-05-02
fix(clean-ignore): Consider .gitignore exclusions

jmgate:
I ran across what I think is the same issue as reported in #76.  My repository's `.gitignore` has, among other things,
```.gitignore
*.mat
!path/to/particular.mat
```
Running `clean-ignore --path-glob "*"` results in all `*.mat` files being stripped from the repository history.  I'm using
```
git version 2.39.3 (Apple Git-146)
```
and if I run
```bash
git check-ignore --verbose --non-matching --no-index path/to/particular.mat
```
I see
```
.gitignore:2:!path/to/particular.mat    path/to/particular.mat
```
It sounds like the expected behavior from `git check-ignore` is to see no output, but that doesn't appear to be the case here.  In order to get the `clean-ignore` script to correctly handle the exclusions in the `.gitignore` file, I needed to make this slight tweak to it.

If it's the case that the real problem lies in `git` itself, and not in this repository, then I could file an issue against it instead.

### DIFF
--- a/contrib/filter-repo-demos/clean-ignore
+++ b/contrib/filter-repo-demos/clean-ignore
@@ -55,8 +55,12 @@ class CheckIgnores:
         if rest == b'::':
           self.okay.add(name)
         else:
-          self.ignored.add(name)
-          ignored.add(name)
+          rule = rest.split(b":")[-1]
+          if rule.decode()[0] == "!":
+            self.okay.add(name)
+          else:
+            self.ignored.add(name)
+            self.add(name)
 
     return ignored
 


newren:
Awesome, thanks for the contribution!  That's another one from Sandia.  :-)  Any chance you know of folks in Engineering Sciences (1500) -- is it still called and numbered that?  I used to work over in that center.

Anyway, I made a couple small tweaks to transform to 6588bbc98ca0 (clean-ignore: handle .gitignore exclusions, 2024-05-02), and merged it in.  Thanks!

jmgate:
Yup, same name and number.  I worked in 1400 for a number of years and had a number of touch points with folks in 1500.

PR #581, opened 2024-07-08
Update INSTALL.md

dhowe:
I believe this is a typo


dhowe:
guess I was wrong as both work in a new shell

Issue #614, opened 2024-11-29
Test fails with zlibng (t9394-filter-repo-sanity-checks-and-bigger-repo-setup.sh)


sebastianas:
I've seen that the main branch has some zlib-ng fixes. Unfortunately it does not work me.
It starts with the size detection which goes wrong. The test string hashes to e80fdf8cd5fb645649c14f41656a076dedc4e12a, too, but the size is reported as 77. This makes zlibng=-5. Based on test, the expected result would be 70.

The next failure is at
`  Total packed size (bytes): $((387+${zlibng}))`

With the size addition this goes to 382 but the files reads as 
`   Total packed size (bytes): 399 `

So the difference isn't constant. However, in blob-shas-and-paths.txt I have
`   a89c82a2d4b713a125a4323d25adda062cc0013d         44         60 numbers/medium.num`
so that is 12 again.

This is true for the remaining checks. I get the test passed if I skip the zlibng detection and hardcode it to 12.
Could we maybe relax the compressed size part? I haven't tested this on anything besides amd64 (zlib-ng_2.2.2) and I am not sure if the results are consistent there but it already appears to be different what you got.

newren:
Yuck, what a pain.

> Could we maybe relax the compressed size part?

Yeah, we'll probably have to do something like that.

Since this only affects the tests and not the functioning of the code, it might be a while before I get to it...

sebastianas:
On 2024-11-29 16:37:08 [-0800], Elijah Newren wrote:
> Yeah, we'll probably have to do something like that.
> 
> Since this only affects the tests and not the functioning of the code,
> it might be a while before I get to it...

I prepared something in case you want to roll with it.

Sebastian

PR #615, opened 2024-11-30
t9394: Mask compressed size for comparison.

sebastianas:
Instead of trying to expect a static difference of the compressed content between zlib and zlib-ng mask the compressed size and ignore it while comparing.

Fixes: #614

### DIFF
--- a/t/t9394-filter-repo-sanity-checks-and-bigger-repo-setup.sh
+++ b/t/t9394-filter-repo-sanity-checks-and-bigger-repo-setup.sh
@@ -486,16 +486,8 @@ test_expect_success C_LOCALE_OUTPUT '--analyze' '
 	(
 		cd analyze_me &&
 
-		# Detect whether zlib or zlib-ng are in use; they give
-		# slightly different compression
-		echo e80fdf8cd5fb645649c14f41656a076dedc4e12a >expect &&
-		python3 -c "print(\"test\\t\" * 1000, end=\"\")" | git hash-object -w --stdin >actual &&
-		test_cmp expect actual &&
-		compressed_size=$(python3 -c "import os; print(os.path.getsize(\".git/objects/e8/0fdf8cd5fb645649c14f41656a076dedc4e12a\"))") &&
-		zlibng=$((72-${compressed_size})) &&
-		test $zlibng -eq "0" -o $zlibng -eq "2" &&
-
-		# Now do the analysis
+		# Do the analysis, mask compressed size away to avoid different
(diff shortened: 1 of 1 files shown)


newren (changes requested):
Ah, don't even try to compare packed sizes at all?  That's kind of a big hammer, I was thinking of putting in some kind of relative tolerance, but it's certainly easier this way.  Maybe that's good enough.

Two things that we'd need fixed, though:
  * You had a different author than committer on your commit.  They either need to be the same, or we need a signoff from both identities.
  * Please drop the "Fixes: #614" trailer.  I'm still trying to make sure the history would make sense for being merged into git, and that trailer wouldn't make sense.

sebastianas:
On 2024-11-30 20:10:21 [-0800], Elijah Newren wrote:
> Ah, don't even try to compare packed sizes at all?  That's kind of a
> big hammer, I was thinking of putting in some kind of relative
> tolerance, but it's certainly easier this way.  Maybe that's good
> enough.

You have cmp in there and I was nore sure how to add something that
accepts a range of values in a specific column. Also the compressed
values are larger than the original - not something one expects :)

> Two things that we'd need fixed, though:
>   * You had a different author than committer on your commit.  They either need to be the same, or we need a signoff from both identities.
>   * Please drop the "Fixes: #614" trailer.  I'm still trying to make sure the history would make sense for being merged into git, and that trailer wouldn't make sense.

Fixed both.

newren:
> You have cmp in there and I was nore sure how to add something that accepts a range of values in a specific column. Also the compressed values are larger than the original - not something one expects :)

Oh, there's no builtin tooling for trying to do range or tolerance comparison.  It'd very much be a case of writing something a script or something for the testsuite specifically for that purpose.

> Fixed both.

Thanks.  Unfortunately, it appears the tests fail on mac, I believe due to the use of `\s` being passed to sed (see https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed).

sebastianas:
On 2024-12-02 00:50:55 [-0800], Elijah Newren wrote:
> Thanks.  Unfortunately, it appears the tests fail on mac, I believe
> due to the use of `\s` being passed to sed (see
> https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed).

Thank you for the hint. I replaced it with [[:space:]] and keep my
fingers crossed :)

Sebastian

newren (approved):

newren:
Thanks!

PR #624, opened 2024-12-25
get_config_settings: Avoid break with multiline configs

ribalda:
If a user has a multiline config like:
git config test.test = "Hellow
world"

git-filter-repo breaks:
>     return dict(line.split(b'=', maxsplit=1)
                  for line in output.strip().split(b"\n"))
E     ValueError: dictionary update sequence element #12 has length 1; 2 is required

Multiline configs are required by b4 in branch-description strategy:
https://b4.docs.kernel.org/en/latest/contributor/prep.html#prep-cover-strategies

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1688,9 +1688,11 @@ class GitUtils(object):
     except subprocess.CalledProcessError as e: # pragma: no cover
       raise SystemExit('fatal: {}'.format(e))
 
+    configs = map(lambda x: x.strip().split(b'=', maxsplit=1), output.splitlines())
+    # FIXME: Ignores multi-line values, just take the first line for now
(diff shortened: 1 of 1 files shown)


scosu:
I stumbled into the same issue with b4, but I didn't see this PR and created another PR for this issue yesterday. I used re.split() but honestly any solution would be fine for me #632

newren:
Thanks for sending this in and sorry for the delay.  I ultimately went with the solution submitted at https://lore.kernel.org/git/20241214181306.296673-1-toke@toke.dk/ and which is now 4697eeb37b7c (Don't crash on multi-line config values, 2024-12-14), as I think it's a bit more robust solution and has a testcase.

Issue #603, opened 2024-10-09
get_config_settings errors on blank lines returned by git config --list


GavinMasterson:
## Issue Description

When running `git filter-repo`, I get an error from the `get_config_settings` function.

## Cause

The output of line 1556's `subproc.check_output(git config --list.split(), cwd=repo_working_dir)` is equivalent to: 
```
output = b'''user.email=name@email.com
user.name=My Name

includeif.gitdir/i:~/work/.path=~/work/.gitconfig-work
user.email=name@workemail.com
user.name=My Name

credential.helper=cache
init.defaultbranch=main
includeif.hasconfig:remote.*.url:git@host.com:work/**.path=~/work/.gitconfig-work
credential.helper=cache
init.defaultbranch=main
core.repositoryformatversion=0
core.filemode=true
core.bare=false
core.logallrefupdates=true'''
```

The conversion of the `output` object to the returned `dict` fails with the message:
> Traceback (most recent call last):
  File "/usr/bin/git-filter-repo", line 4151, in <module>
    main()
  File "/usr/bin/git-filter-repo", line 4148, in main
    filter.run()
  File "/usr/bin/git-filter-repo", line 4068, in run
    self._run_sanity_checks()
  File "/usr/bin/git-filter-repo", line 2931, in _run_sanity_checks
    self._config_settings = GitUtils.get_config_settings(target_working_dir)
  File "/usr/bin/git-filter-repo", line 1563, in get_config_settings
    return dict(line.split(b'=', maxsplit=1)
ValueError: dictionary update sequence element #2 has length 1; 2 is required

## Potential Solution

I fixed the issue by modifying the function as follows: 
```
@staticmethod
  def get_config_settings(repo_working_dir):
    output = ''
    try:
      output = subproc.check_output('git config --list'.split(),
                                    cwd=repo_working_dir)
    except subprocess.CalledProcessError as e: # pragma: no cover
      raise SystemExit('fatal: {}'.format(e))
    
    # Changes start here -->
    decoded_output = output.decode('utf-8')
    lines = [line for line in decoded_output.split('\n') if line.strip()]
    # FIXME: Ignores multi-valued keys, just let them overwrite for now
    return dict(line.split('=', maxsplit=1) for line in lines if '=' in line)
``` 

There are probably better ways to solve the issue that do not involve an opinionated decoding step, as I cannot tell what side-effects the decoding step may have in repos that are not UTF-8 encoded.

## Summary

[...]

PR #632, opened 2025-01-21
filter-repo: Accept multiline git config values

scosu:
Some git config options can contain multiline values, for example aliases. With the current simple split on = this fails with a stacktrace so that git-filter-repo is unusable.

This patch uses a multiline regular expression to split the git config --list string at any beginning line followed by an alphanumeric name with '-' and '.' followed by '='. In some cases this can falsely identify multiline values as a new variable name that does not exist. But it should work for most usual multiline config options correctly.

Corresponding bugreport #603

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -1689,8 +1689,9 @@ class GitUtils(object):
       raise SystemExit('fatal: {}'.format(e))
 
     # FIXME: Ignores multi-valued keys, just let them overwrite for now
-    return dict(line.split(b'=', maxsplit=1)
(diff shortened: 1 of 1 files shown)


newren:
Thanks for sending this in.  I ultimately went with the solution submitted at https://lore.kernel.org/git/20241214181306.296673-1-toke@toke.dk/ and which is now 4697eeb37b7c (Don't crash on multi-line config values, 2024-12-14), as I think it's a bit more robust solution and has a testcase.

PR #637, opened 2025-01-29
Give a more succinct error message when user specifies nonexistent files

cincodenada:
Currently specifying nonexistent files to options gives a whole Python backtrace, with the filename output in `b'filename'` bytestring notation which is even more confusing.

There are plenty of approaches here, this just catches any FileNotFoundErrors and raises a more succinct error message about it, which doesn't tell you which is the offending argument, but that should be apparent enough, and in any case, this seems to me like strictly an improvement over the backtrace that currently happens.

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -4965,7 +4965,12 @@ class RepoFilter(object):
 
 def main():
   setup_gettext()
-  args = FilteringOptions.parse_args(sys.argv[1:])
+  try:
+    args = FilteringOptions.parse_args(sys.argv[1:])
+  except FileNotFoundError as e:
+    filename = e.filename.decode('utf-8') if type(e.filename) == bytes else e.filename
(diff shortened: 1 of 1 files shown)


newren (changes requested):
Hi,

Thanks for sending this in.  The general idea seems good, but this particular placement feels a bit to coarse to me; I'd rather have the checks around get_replace_text() and the usage of args.strip_blobs_with_ids.

Also, the commit message mostly looks good, but to follow the git style the first line should be "area: summary", e.g. "filter-repo: handle nonexistent files more nicely" (also notice the lack of capitalization after the colon).  It may also be worth adding in the text after the first line that this is specifically about argument parsing, but the main thing is just the "area: summary" bit.

newren:
It's been a while with no response; I'll go ahead and close out.  Feel free to reopen if you want to adjust according to my comment above, though.

Issue #640, opened 2025-02-05
Missing line-break after message "Warning: Removing recursively:"


goran-w:
Issue: The following print-out message is missing a terminating line-break and gets jumbled with other text:
`Warning: Removing recursively: <filename>`

Location in source-code: https://github.com/newren/git-filter-repo/blob/main/git-filter-repo#L2906

Example output:
```
Processed 28217 blob sizesely: ".git\filter-repo\analysis"
Processed 1129 commits
Writing reports to .git\filter-repo\analysis...done.
```
(Notice how the words `sizes` and `recursively` get combined into `sizesely` on the first line, due to repeatedly overwritten runtime progress reports... This would be solved by adding the missing line-break.)

Adding the missing newline will correct the output:
```
Warning: Removing recursively: ".git\filter-repo\analysis"
Processed 28217 blob sizes
Processed 1129 commits
Writing reports to .git\filter-repo\analysis...done.
```

PR #644, opened 2025-02-19
Add missing newline in printout

goran-w:
Fixes the issue reported in #640.

This print-out message is missing a terminating line-break,
and gets partially overwritten by other text:

Warning: Removing recursively: <filename>

Example output:

Processed 28217 blob sizesely: ".git\filter-repo\analysis"

(Notice how the words "sizes" and "recursively" get combined,
into "sizesely" on the first line.)

Adding the missing newline corrects the output:

Warning: Removing recursively: ".git\filter-repo\analysis"
Processed 28217 blob sizes

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -2903,7 +2903,7 @@ class RepoAnalyze(object):
 
     if os.path.isdir(reportdir):
       if args.force:
-        sys.stdout.write(_("Warning: Removing recursively: \"%s\"") % decode(reportdir))
+        sys.stdout.write(_("Warning: Removing recursively: \"%s\"\n") % decode(reportdir))
         shutil.rmtree(reportdir)
       else:
         sys.stdout.write(_("Error: dir already exists (use --force to delete): \"%s\"\n") % decode(reportdir))


newren (changes requested):
The code modification looks fine, but there's some issues with the commit:
  * I need you to use a real email address, not @users.noreply.github.com
  * The committer needs to likewise be a real person, not `GitHub <noreply@github.com>` (I suspect that means you used GitHub online editing; don't.)
  * You need to add a Signed-off-by line
  * Please use imperative tense (`add` vs. `added`)
  * The first line should be of the form `area: summary`, e.g. `filter-repo: add missing newline`
  * The commit message should stand on its own; please add extra information in the commit message and remove the `(#640)` reference.

newren (changes requested):
Hi,

Thanks for fixing this up; looks better, though there's still a few issues:

* I don't want to implicitly depend upon the project being hosted at github.com for it to make sense; it should be allowed to be moved elsewhere.  The "#640" reference only makes sense if people know that implicit mapping.  We could strike the parenthetical comment, but then the remainder of the line is somewhat duplicative of the line after it.  Can we just remove the whole line with "#640" on it?
* Please wrap the long lines in the commit message at 72 characters
* The Author, Committer, and Signed-off-by need to be a real name, not a username.

If you can fix these up, I'd be happy to merge.

newren:
Thanks!

PR #646, opened 2025-02-27
Pr/log readme

dimateos:
I saw the output dir in the log, but could not control+click due to a missing space:
(underline appears on hover on clickable paths/links)

image

With this PR, I added the space plus a bit formatting.
I also added the readme file because opening files with control+click is more convenient than folders IMO.

image

Demo showing the control+click in vscode:

2025-02-27_(082159)_Code

I split the commits along with detailed descriptions, but I can squash them or whatever to may need
Best regards!

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -2903,7 +2903,7 @@ class RepoAnalyze(object):
 
     if os.path.isdir(reportdir):
       if args.force:
-        sys.stdout.write(_("Warning: Removing recursively: \"%s\"") % decode(reportdir))
+        sys.stdout.write(_("Warning: Removing recursively: \"%s\"\n") % decode(reportdir))
         shutil.rmtree(reportdir)
       else:
         sys.stdout.write(_("Error: dir already exists (use --force to delete): \"%s\"\n") % decode(reportdir))


git-filter-repo
```diff
@@ -2903,10 +2903,10 @@ class RepoAnalyze(object):
 
     if os.path.isdir(reportdir):
       if args.force:
-        sys.stdout.write(_("Warning: Removing recursively: \"%s\"\n") % decode(reportdir))
+        sys.stdout.write(_("WARNING: Removing recursively: \"%s\"\n") % decode(reportdir))
         shutil.rmtree(reportdir)
       else:
-        sys.stdout.write(_("Error: dir already exists (use --force to delete): \"%s\"\n") % decode(reportdir))
+        sys.stdout.write(_("ERROR: dir already exists (use --force to delete): \"%s\"\n") % decode(reportdir))
```
newren, line 2909: We have 29 messages starting with "Error", and none starting with "ERROR" in the current codebase.  I think it feels weird to change 1 of them.

"Warning" is a bit different.  We only have three, with one being all-caps.  But, for consistency, I'd say we should probably change the all-caps one to a single capital, and then not make this change.

git-filter-repo
```diff
@@ -2915,10 +2915,10 @@ class RepoAnalyze(object):
     stats = RepoAnalyze.gather_data(args)
 
     # Write the reports
-    sys.stdout.write(_("Writing reports to %s...") % decode(reportdir))
+    sys.stdout.write(_("Writing reports to \"%s\"...") % decode(reportdir))
```
newren, line 2918: This is reasonable.

git-filter-repo
```diff
@@ -2915,10 +2915,11 @@ class RepoAnalyze(object):
...
     sys.stdout.flush()
     RepoAnalyze.write_report(reportdir, stats)
-    sys.stdout.write(_("done.\n"))
+    sys.stdout.write(_(" done.\n"))
```
newren, line 2921: I don't understand the point of adding the space.  Your commit message says it's to make the filename ctrl-clickable I think, but if that were the case then you'd need the space before the ellipsis, right?

git-filter-repo
```diff
@@ -2915,10 +2915,11 @@ class RepoAnalyze(object):
...
+    sys.stdout.write(_("README: \"%s\"\n") % decode( os.path.join(reportdir, b"README") ))
```
newren, line 2922: Not sure whether this belongs more as a parenthetical on the line where we mention writing reports, or if this is fine.  I guess it's okay as-is.

newren (changes requested):

git-filter-repo
dimateos: At least with vscode, ellipsis at the end of a path are ok. So I kept it and simply spaced the "done" from the ellipsis.
image

In the GIF attached to the PR you can see both paths being underlined correctly on hover.
The quotes improved compatibility, so actually the space is not required with them.

git-filter-repo
dimateos: As you wish, I used two separate lines to prevent paths getting cut due to terminal width.
Seems more compatible to avoid breaking ctrl+clicking, but in fact vscode can handle it (don't know other editors etc):
image

git-filter-repo
newren: Great, since the quotes solved the issue and the extraneous space isn't necessary, let's toss it.

git-filter-repo
```diff
@@ -4455,7 +4456,7 @@ class RepoFilter(object):
       ret = subproc.call([arg if arg != '""' else '' for arg in cmd.split()],
                          cwd=source_working_dir)
       if ret != 0: # pragma: no cover
-        m = _("WARNING: Fetching all refs from origin failed")
+        m = _("Warning: Fetching all refs from origin failed")
```
newren, line 4459: I appreciate the desire to make things consistent, but it's unrelated to your other changes in this PR and should probably be separate.

newren (changes requested):
In addition to the comments on individual lines...rather than have 3 commits where one commit primarily just undoes the change of the other, let's make the history clean

git-filter-repo
dimateos: Sure, you want me to drop it or adjust the description (PR / commit)?
I edited it so the commit was not simply a revert but apply the potential change you mentioned.

Let me know and I will squash the changes into a single one so you can review it :)

newren: I'm saying that commit 3 is mostly a revert of about half of commit 2.  The commit(s) should tell what is roughly the simplest way to get from the starting state to the end state, and doing and then undoing changes is rarely the simplest path to the end state.  The PR description should describe the overall state and goals of the commits, which should have a coherent theme (which seems to me to be tweaks to the log messages for --analyze).  The parts of commit 3 that would remain relevant don't quite fit into that theme to me, so saving that for a different PR made more sense to me.  You might be able to persuade me otherwise on that point, but the do-some-changes-then-undo definitely should be cleaned up; more important than the PR description to me are the commits since they will become part of the history unlike the PR description.

dimateos: I agree with it better being a single commit, and as I said the intention is to squash them into a single commit.
Will do so and remove the edit on the unrelated all-caps warning message, introduced in commit 3.

dimateos (commented):
> Changes were reviewed (not sure why there are some still pending, now outdated)

newren (changes requested):
Hi, sorry for the delay.

This version looks pretty good, but the commit message needs some touching up:
  * Commit messages should use imperative tense, same as with the git project
  * Lines should be wrapped at 72 characters.

So, the lines before the Signed-off-by would look something like:
```
filter-repo: log README file from reports directory

Make the reports README file more discoverable by logging it.  Further,
to allow control+click opening of the file inside certain
terminals/IDEs, add quotes around the filename.  Add quotes around the
directory that is already displayed as well, for the same reason.
```

dimateos:
Hello!

Thanks for the response, I really appreciate the effort.
I kept the message as you put it as it seemed to perfectly describe the change.

Cheers!

PR #657, opened 2025-04-15
Avoid extra commit id changes by respecting Git tree ordering

Beuc:
A no-op test rewrite of git-filter-repo, using `git-filter-repo --proceed`, unexpectedly results in a different tree.

Diffing the fast-export dumps with `--dry-run`, it appears git-filter-repo is using Python's default sorting algorithm for Git trees, rather than the Git-specific algorithm, which is described a bit at: https://stackoverflow.com/questions/78981525/how-git-sort-entries-in-trees

```
@@ -1451,25 +1329,23 @@
 D testcases/expected/case1-twenty
 D testcases/inputs/case1
 M 100755 0a13abf13c87a13ae56c1fc664baa8b44bdac1de testcases/t9390-repo-filter.sh
+M 100644 de3799fa8cd84a9ac56fad023ddf3da44eb444bb testcases/t9390/case1
 M 100644 e0c88454095bd887852791401a3b2336ad47d012 testcases/t9390/case1-filename
 M 100644 a1aa78ff11f3a40cafdcd9b531e8c1342e8a75e5 testcases/t9390/case1-ten
 M 100644 488cbd9d101b107809611940d1cd00a8f72e1b17 testcases/t9390/case1-twenty
-M 100644 de3799fa8cd84a9ac56fad023ddf3da44eb444bb testcases/t9390/case1
```

Unlike described above though, it appears the fast-export dumps treats all files as directory when sorting them (`case1` is a file in the example above), so appending "/" to all entries maintains the original tree order.

This avoids unnecessary tree hash variation and early, cascading commit id changes, which shouldn't happen before a commit is effectively rewritten by git-filter-repo.

With this change, `git-filter-repo --proceed` is stable and produces an identical repository.


newren (changes requested):
Hi,

You'll note that when you submitted this build, the "DCO" had a red X by it.  Clicking on the "view details" next to it would explain how the commit needs a Signed-off-by trailer to pass.

The reference to "sorting order" in your commit message is actually misleading, because fast-export and fast-import don't operate on trees.  In fact, fast-export isn't using git's normal sort algorithm; it has a fast-export-specific sort (https://github.com/git/git/blob/14de3eb34435db79c6e7edc8082c302a26a8330a/builtin/fast-export.c#L444-L448) which doesn't match what your code does.  (It'll match in more cases than python's default sort, but you're still sorting differently.)  So you'll want to fix that up.

The author on the commit appears broken.  The committer looks fine (view both with `git log --pretty=fuller`).  Can you fix up the author?  (`git commit --reset-author --amend` ought to do it).

Commit messages should also be wrapped at 72 characters, unless you're quoting output of some command.

The first line of commit messages should generally be of the form "area: brief change" rather than just "break change".

Since that's a lot of comments I made on the commit message, I'll provide an example below:
```
filter-repo: reduce export dump differences

While the order of files passed to `git fast-import` do not matter, we
can help users inspect --dry-run results if we can match the order that
fast-export uses.  fast-export uses a depth-first sorting order to
ensure that entries underneath a directory would appear before the
directory or some kind of file or symlink that has replaced a directory.
However, its depth-first ordering is implemented in a way that affects
things besides directories becoming symlinks or files -- in contrast to
normal lexicographic ordering where a substring would sort before a
longer string, fast-export reverses that.  This means that when we try
to inspect differences after a --dry-run, we may see a spurious
difference such as:

    $ diff -u .git/filter-repo/fast-export.original .git/filter-repo/fast-export.filtered
    ...
    @@ -1451,25 +1329,23 @@
     D testcases/expected/case1-twenty
     D testcases/inputs/case1
     M 100755 0a13abf13c87a13ae56c1fc664baa8b44bdac1de testcases/t9390-repo-filter.sh
    +M 100644 de3799fa8cd84a9ac56fad023ddf3da44eb444bb testcases/t9390/case1
     M 100644 e0c88454095bd887852791401a3b2336ad47d012 testcases/t9390/case1-filename
     M 100644 a1aa78ff11f3a40cafdcd9b531e8c1342e8a75e5 testcases/t9390/case1-ten
     M 100644 488cbd9d101b107809611940d1cd00a8f72e1b17 testcases/t9390/case1-twenty
    -M 100644 de3799fa8cd84a9ac56fad023ddf3da44eb444bb testcases/t9390/case1

Fix this by modifying filter-repo to use the same sorting of paths as
fast-export when --dry-run is specified.

Signed-off-by: FILL THIS OUT
```

Beuc:
Hi.

I rewrote the patch with a `depth_first`-compatible sort algorithm.
I also reused your provided description (I only added the sample `git-filter-repo` command used to generate the diffs).

(Note: while testing, I found a quite simpler patch that we might consider:)
```
         continue
       # Otherwise, record the change
       new_file_changes[change.filename] = change
-    commit.file_changes = [v for k,v in sorted(new_file_changes.items())]
+    commit.file_changes = [v for k,v in new_file_changes.items()]
 
   def _tweak_commit(self, commit, aux_info):
     if self._args.replace_message:

```

Anyway this should be good now.

newren:
> (Note: while testing, I found a quite simpler patch that we might consider:)
> 
> ```
>          continue
>        # Otherwise, record the change
>        new_file_changes[change.filename] = change
> -    commit.file_changes = [v for k,v in sorted(new_file_changes.items())]
> +    commit.file_changes = [v for k,v in new_file_changes.items()]

Ooh, that'd be much simpler and faster; let's do that!

PR #668, opened 2025-05-21
Postpone refs matching to after setting working dir to repo dir.

corto-pabu:
subproc.Popen() calls in bfg-ish relies on a os.chdir() call to have been made beforehand for each child process to run in the repo dir as required.

get_preservation_info has a subproc.Popen() call that before this change is executed before the required os.chdir() call.

### DIFF
--- a/contrib/filter-repo-demos/bfg-ish
+++ b/contrib/filter-repo-demos/bfg-ish
@@ -382,10 +382,11 @@ class BFG_ish:
 
   def run(self):
     bfg_args = self.parse_options()
-    preserve_refs = self.get_preservation_info(bfg_args.preserve_ref_tips)
 
     work_dir = os.getcwd()
     os.chdir(bfg_args.repo)
+
+    preserve_refs = self.get_preservation_info(bfg_args.preserve_ref_tips)
     bfg_args.delete_files = java_to_fnmatch_glob(bfg_args.delete_files)
     bfg_args.delete_folders = java_to_fnmatch_glob(bfg_args.delete_folders)
     bfg_args.filter_content_including = \


newren (changes requested):
Nice catch and well spotted.

The modification itself looks good, but there's a few issues with the commit:
  * All commits need a Signed-off-by trailer for me to accept them, same as required by the git project.
  * Please wrap lines in the commit message at 72 characters
  * First lines should be of the form "area: change summary" rather than just "change summary".
  * I need a real committer rather than "GitHub <noreply@github.com>"; that'll probably be resolved automatically by amending your commit to add your signed-off-by, though.
 
Taking this all together, you probably just need to run `git commit --amend -s` on this commit and use a commit message that looks something like:

```
bfg-ish: postpone refs matching until working dir has been adjusted

subproc.Popen() calls in bfg-ish rely on an os.chdir() call to have been
made beforehand so that the child process runs in the repo directory as
required.  Unfortunately, get_preservation_info() has a subproc.Popen()
call before the os.chdir(), which is problematic.  Move the
get_preservation_info() call after the os.chdir() call.

Signed-off-by: <FILL THIS OUT>
```

If you can fix that up, I'd be happy to merge it.

Rahul-cyber26 (commented):
reassigns the preserve_refs variable after changing to the repository directory (os.chdir(bfg_args.repo)). This ensures that preservation info is obtained based on the current working directory, which may be necessary if the method depends on files or context within the repository.

phillipschandler19-web:
is ahead, behind or both.

$ git branch -vv
  iss53     7e424c3 [origin/iss53: ahead 2] Add forgotten brackets
  master    1ae2a45 [origin/master] Deploy index fix
* serverfix f8674d9 [teamone/server-fix-good: ahead 3, behind 1] This should do it
  testing   5ea463a Try something new$ git rebase --onto master server clientINSTALL.md

phillipschandler19-web:
$ git checkout master
Switched to branch 'master'

$ vim hello.rb
$ git diff
diff --git a/hello.rb b/hello.rb
index ac51efd..36c06c8 100755
--- a/hello.rb
+++ b/hello.rb
@@ -1,5 +1,6 @@
 #! /usr/bin/env ruby

+# prints out a greeting
 def hello
   puts 'hello world'
 end

$ git commit -am 'Add comment documenting the function'
[master bec6336] Add comment documenting the function
 1 file changed, 1 insertion(+)

PR #720, opened 2025-11-10
Add rough cut of 'pathbase'

studog-github:
This is a rough draft for a "pathbase" feature. It operates like git's interactive rebase but is not interactive. I'm not intending for this to be merged; just looking for comments on whether this could be merged. It would need test cases at least.

After running `git-filter-repo --analyze --one-path-per-blob`, one annotates the resulting `.git/filter-repo/analysis/blob-shas-and-paths.txt` with commands and feeds that back in via `git-filter-repo --pathbase`. This avoids most of the human error inherent in a person transforming a `blob-shas-and-paths.txt` into a file that `--paths-from-file` can handle.

It is a process that most git users will already be familiar with from rebase.

Unless the extra line in the `blob-shas-and-paths.txt` file causes issues, this is 100% backwards compatible.

I am currently using this to clean a repo that had a lot of build artifacts checked in. I find it better than (re-)creating a pathnames-only file from `blob-shas-and-paths.txt` because I can see what I'm selecting and it's less transformation than making the pathnames-only file.
Part of my equation is this lets me do the annotation slowly over time/many iterations while I figure out what needs to come out and what stays.

Yes, this is a different approach than the "blobbase" PR I posted. It turns out that in a lot of cases, I want to remove an item from an erroneously checked in build products directory, but that item is an art/etc asset that should still reside in the codebase. Thus selecting the sha for stripping removes both, and I only want one.

Also, the 0-byte deduplication blob sha has a lot of unrelated files pointing to it; those also have to be filtered by path.

### DIFF
--- a/git-filter-repo
+++ b/git-filter-repo
@@ -43,6 +43,7 @@ import subprocess
 import sys
 import time
 import textwrap
+import re
 
 from datetime import tzinfo, timedelta, datetime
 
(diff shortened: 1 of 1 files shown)


Sawsqr68:
[Run tests](https://github.com/newren/git-filter-repo/actions/workflows/test.yml)

PR #725, opened 2025-12-09
Avoid wrapping git commands on newlines

SamuelT-Beslogic:
There's an entire section dedicated to avoiding dumb, easy to make mistakes, yets wraps the command to copy with a newline, which would cause a user to paste an incomplete command in their terminal and send before they have time to validate.

This small documentation change aims to help protect users against themselves.


SamuelT-Beslogic:
Also as a side note, out of scope here, but
```shell
> git push --force --branches --tags --prune
error: unknown option `branches'
```

PR #733, opened 2026-01-18
fix broken link syntax

electricatgit:
just a guess following the convention I see elsewhere in the file.

### DIFF
--- a/Documentation/git-filter-repo.txt
+++ b/Documentation/git-filter-repo.txt
@@ -290,7 +290,7 @@ Generic callback code snippets
 
 --file-info-callback <function_body>::
 	Python code body for processing the combination of filename, mode,
-	and associated file contents; see <<CALLBACKS>.  Note that when
+	and associated file contents; see <<CALLBACKS>>.  Note that when
 	--file-info-callback is specified, any replacements specified by
 	--replace-text will not be automatically applied; instead, you
 	have control within the --file-info-callback to choose which files


chriscool:
Thanks for your PR. I think the following could be improved though:

- commit message subject:

Something like "git-filter-repo.txt: fix broken link syntax" or even "docs: fix broken link syntax" would be better than just "fix broken link syntax".

- full name and email in the `Signed-off-by:` trailer:

The Git Documentation at https://git-scm.com/docs/SubmittingPatches#sign-off says:

```
Please use a known identity in the `Signed-off-by` trailer, since we cannot
accept anonymous contributions. It is common, but not required, to use some form
of your real name. We realize that some contributors are not comfortable doing
so or prefer to contribute under a pseudonym or preferred name and we can accept
your patch either way, as long as the name and email you use are distinctive,
identifying, and not misleading.

The goal of this policy is to allow us to have sufficient information to contact
you if questions arise about your contribution.
```

while your Signed-off-by trailer is:

`Signed-off-by: electricat <git.devnull@electricat.mayfirst.org>`

which I am afraid doesn't give away a lot of information about you.  The `devnull` part in your email address is also not very encouraging. So not sure we have sufficient information to contact you if questions arise about your contribution.

I guess a typo fix like this is not really copyrightable, so it's not likely that there will be any question about this contribution. But anyway to keep the list of contributors clean, it would not be nice to accept a contribution with such a Signed-off-by.

Would you mind changing this to something more in line with the Git documentation? Otherwise would you accept that I, or someone else, take over your contribution?

Thanks!

chriscool:
I have fixed the broken link in

https://github.com/newren/git-filter-repo/pull/752

where I have added a "Suggested-by: ..." trailer to credit you as well as my "Signed-off-by: ..." trailer.

Thanks again for this PR.

Now I think it should be closed in favor of https://github.com/newren/git-filter-repo/pull/752

Issue #559, opened 2024-04-30
Support for SHA256


james2432:
As of git 2.45, SHA256 is no longer considered experimental. SHA256 seems like the way forward and should eventually be the ultimate goal

https://lore.kernel.org/git/xmqq8r0ww0sj.fsf@gitster.g/T/#u

https://github.blog/2024-04-29-highlights-from-git-2-45/

newren:
Not sure whether to label this as a bug or enhancement, but yeah, this is important.

emzxcv:
> Not sure whether to label this as a bug or enhancement, but yeah, this is important.

Any update on this? We would love to have this worked on.

PR #736, opened 2026-02-15
SHA256 support

lel-amri:
DRAFT


lel-amri:
Attempt to fix #559. WIP (this comment will be rewritten). Missing tests. Have to clean some parsing done prior to object format detection.

I went for detection of object format from the `git fast-export` output. This is the hardest way. There is an easier one that should also yield cleaner code that would be to check repo's `objectFormat` config value prior to parsing, and thus set the parser to the expected object format at initialization. What do you think @newren?

phillipschandler19-web:
INSTALL.md$ git checkout master
Switched to branch 'master'

$ vim hello.rb
$ git diff
diff --git a/hello.rb b/hello.rb
index ac51efd..36c06c8 100755
--- a/hello.rb
+++ b/hello.rb
@@ -1,5 +1,6 @@
 #! /usr/bin/env ruby

+# prints out a greeting
 def hello
   puts 'hello world'
 end

$ git commit -am 'Add comment documenting the function'
[master bec6336] Add comment documenting the function
 1 file changed, 1 insertion(+)git checkout -b whitespace
Switched to a new branch 'whitespace'

$ unix2dos hello.rb
unix2dos: converting file hello.rb to DOS format ...
$ git commit -am 'Convert hello.rb to DOS'
[whitespace 3270f76] Convert hello.rb to DOS
 1 file changed, 7 insertions(+), 7 deletions(-)

$ vim hello.rb
$ git diff -b
diff --git a/hello.rb b/hello.rb
index ac51efd..e85207e 100755
--- a/hello.rb
+++ b/hello.rb
@@ -1,7 +1,7 @@
 #! /usr/bin/env ruby

 def hello
-  puts 'hello world'
+  puts 'hello mundo'^M
 end

 hello()

$ git commit -am 'Use Spanish instead of English'
[whitespace 6d338d2] Use Spanish instead of English
 1 file changed, 1 insertion(+), 1 deletion(-)
Now we switch back to our master branch and add$ git cat-file -p HEAD
tree cfda3bf379e4f8dba8717dee55aab78aef7f4daf
author Scott Chacon  1301511835 -0700
committer Scott Chacon  1301511835 -0700

initial commit

$ git ls-tree -r HEAD
100644 blob a906cb2a4a904a152...   README
100644 blob 8f94139338f9404f2...   Rakefile
040000 tree 99f1a6d12cb4b6f19...   lib

PR #743, opened 2026-05-17
Add SLSA generic generator workflow

njegus:
This workflow generates SLSA provenance files for projects, satisfying level 3 requirements. It includes steps for building artifacts and generating provenance subjects.


newren:
Stupid spam accounts.

